PR #6, opened 2018-09-10
Add missing .com in support email

LightningK0ala:
During initial setup, in the bag number check screen, the support email is missing .com.


peter-conalgo:
Thanks for this @LightningK0ala but the omission was intentional.

There isn't quite enough room on a single line for the whole email address without word wrap.

This is what it would look like with your change:

snapshot-253-131343

Compared to the original:

snapshot-253-131627

PR #8, opened 2018-10-25
Ui copy edit

toadlyBroodle:
Copy edited the UI strings in /shared/actions.py to fix minor grammatical mistakes, typos, etc.

Love the product btw ;)


peter-conalgo:
The problem is all of this copy was designed for, and carefully tested on the emulator or a real system, and in many cases what looks like a blank line in the python source code is really there to make the message more readable on the device.

I would rather see a series of smaller pull requests for legit errors that are indisputable. Word choice, and awkward phrasing doesn't count.

toadlyBroodle:
Your call. I know how much work you guys have put into the backend. Imo, your UI does not reflect that.

PR #9, opened 2018-11-21
Payment addresses shown in public.txt

vesparny:
please also note that when you say:

> The first few receive addresses should match the values shown in public.txt beside the text: m/44'/0'/0'/0/0, m/44'/0'/0'/0/1 and so on.

I checked and the addresses shown in electrum are different form the ones in the `public.txt`

### DIFF
--- a/docs/electrum-usage.md
+++ b/docs/electrum-usage.md
@@ -54,7 +54,7 @@ have the seed words.
 - On the Coldcard, choose `Dump Summary` from the Advanced menu, and open the
         `public.txt` file that generated on your computer.
 - In `public.txt`, look for the section labeled "BIP44 / Electrum" and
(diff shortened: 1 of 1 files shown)


vesparny:
@nvk ping

peter-conalgo:
Sorry for the delay. Please check that you are comparing the addresses in the appropriate section labeled either:

    For Electrum (not BIP44): m/{change}/{idx}

or

    For BIP44 / Electrum: m/44'/1'/{account}'/{change}/{idx}

Also note, the first `/1'/` will be zero for mainnet, and most examples are usually from testnet, because we use the simulator.

peter-conalgo:
Sorry just looked at the diff now (thought this was an issue). You're correct. Merged.

PR #12, opened 2018-11-27
Document unix socket address format when on linux

achow101:
Linux and MacOS use different formats apparently.

It would be better if a switch or autodetection could be used to switch between formats, but I don't know how to do that here. The format used comes from https://github.com/micropython/micropython/issues/2464#issuecomment-250452170


nvk:
Thanks.

PR #15, opened 2019-05-12
Pass previously chosen word target when starting over

duck1123:
When choosing to start over, the target word length was being reset and causing an error. This change passes the target word count from the previous attempt.


peter-conalgo:
Could you add a `num_words=` in there and then I'll merge.

Thanks for the fix!

duck1123:
done. Also forgot my signature. :)

PR #19, opened 2019-09-22
Add address explorer menu

hodlwave:
Adds a menu for viewing the addresses controlled by the
wallet currently loaded on the ColdCard. The first stage
of the menu prompts the user for the wallet they use
out of a list of common derivation paths. This allows
the user to drill-down into an explorer of their
addresses, which they can write to the MicroSD card.

This feature is useful to users in confirming
receive addresses that they send bitcoin to.


hodlwave:
Once a method to generate redeem scripts is added the MultisigWallet class in `shared/multisig.py`, I can refactor the code so the multisig wallets loaded on device are also included in the Address Explorer menu.

peter-conalgo:
See #20

PR #23, opened 2019-10-15
Add Syscoin support

sidhujag:
Create Syscoin definition.

HWI: https://github.com/syscoin/hwi
Slip-0044: https://github.com/satoshilabs/slips/blob/master/slip-0044.md
Slip-132: https://github.com/satoshilabs/slips/blob/master/slip-0132.md

SYS BIP32 path is m/84' or 44' or 49'/57'/0/0

We use BIP84 in our apps/wallets and segwit by default.

### DIFF
--- a/shared/chains.py
+++ b/shared/chains.py
@@ -261,6 +261,28 @@ class LitecoinTestnet(LitecoinMain):
     # ?? unknown ??
     b44_cointype = 2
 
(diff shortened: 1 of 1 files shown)


nvk:
Thanks for the submission, but we don't intend to include alt coins in the reference Coldcard implementation.

PR #25, opened 2019-10-24
Implement multisig address explorer

hodlwave:
This change required several steps:
- Add methods to create a redeem script and multisig address
- Refactor Address Explorer to use injected get_address_at_idx method
- Implement multisig address explorer menu
- Test for multisig address explorer

The test also required
- adding a fixture in conftest.py to select the blockchain on the device
- refactoring the make_multisig fixture to use multiple chains / addr_fmts

N.B. In order to generalize the make_multisig fixture, I set the last signer to be the simulator's actual extended private key as opposed to the one in the constants file. I don't believe this should have any backwards compatibility issues, but I wanted to make a note of it in case tests start breaking.

### DIFF
--- a/shared/address_explorer.py
+++ b/shared/address_explorer.py
@@ -68,7 +68,7 @@ async def choose_first_address(*a):
 
     return (path, addr_fmt)
 
-async def show_n_addresses(start, n, get_address_at_idx):
+async def show_n_addresses(start, n, get_addresses):
     # Displays n addresses from start
     from main import dis
 
@@ -80,10 +80,9 @@ async def show_n_addresses(start, n, get_address_at_idx):
 
         dis.fullscreen('Wait...')
 
-        for idx in range(start, start + n):
-            subpath, address = get_address_at_idx(idx)
+        for i, (subpath, address) in enumerate(get_addresses(start, start + n - 1)):
             msg += "%s =>\n%s\n\n" % (subpath, address)
-
+            idx = i + start
             dis.progress_bar_show(idx/n)
 
         msg += "Press 9 to see next group, 7 to go back. X to quit."
@@ -97,7 +96,7 @@ async def show_n_addresses(start, n, get_address_at_idx):
 
             if ch == '1':
                 # save addresses to MicroSD signal
-                await make_address_summary_file(get_address_at_idx)
+                await make_address_summary_file(get_addresses)
                 # .. continue on same screen in case they want to write to multiple cards
 
             if ch == 'x':
@@ -113,18 +112,17 @@ async def show_n_addresses(start, n, get_address_at_idx):
                 start += n
                 msg = make_msg(start)
 
-def generate_address_csv(get_address_at_idx, n):
+def generate_address_csv(get_addresses, n):
     # Produce CSV file contents as a generator
 
     yield '"Index","Payment Address","Derivation"\n'
 
     ch = chains.current_chain()
 
-    for idx in range(n):
-        subpath, address = get_address_at_idx(idx)
+    for idx, (subpath, address) in enumerate(get_addresses(0, n-1)):
         yield '%d,"%s","%s"\n' % (idx, address, subpath)
 
-async def make_address_summary_file(get_address_at_idx, fname_pattern='addresses.txt'):
+async def make_address_summary_file(get_addresses, fname_pattern='addresses.txt'):
     # write addresses into a text file on the MicroSD
     from main import dis
     from files import CardSlot, CardMissingError
@@ -137,7 +135,7 @@ async def make_address_summary_file(get_address_at_idx, fname_pattern='addresses
     dis.fullscreen('Saving 0-%d' % count)
 
     # generator function
-    body = generate_address_csv(get_address_at_idx, count)
+    body = generate_address_csv(get_addresses, count)
 
     # pick filename and write
     try:
@@ -188,14 +186,16 @@ Press 4 to start.''', escape='4')
 
     path, addr_fmt = picked
     chain = chains.current_chain()
-    def get_address_at_idx(idx):
+    def get_addresses(first, last=None):
+        if last == None:
+            last = first
         with stash.SensitiveValues() as sv:
-            subpath = path.format(account=0, change=0, idx=idx)
-            node = sv.derive_path(subpath, register=False)
-            res = (subpath, chain.address(node, addr_fmt))
+            for idx in range(first, last + 1):
+                subpath = path.format(account=0, change=0, idx=idx)
+                node = sv.derive_path(subpath, register=False)
+                yield (subpath, chain.address(node, addr_fmt))
             stash.blank_object(node)
-            return res
 
-    await show_n_addresses(0, 10, get_address_at_idx)
+    await show_n_addresses(0, 10, get_addresses)
 
 # EOF
--- a/shared/address_explorer.py
+++ b/shared/address_explorer.py
@@ -68,7 +68,7 @@ async def choose_first_address(*a):
 
     return (path, addr_fmt)
 
-async def show_n_addresses(path, addr_fmt, start, n):
+async def show_n_addresses(start, n, get_address_at_idx):
     # Displays n addresses from start
     from main import dis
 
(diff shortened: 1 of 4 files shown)


hodlwave:
@peter-conalgo @nvk 

Also I held off on adding the multisig wallets to the regular 'Address Explorer' menu for now because I thought the PR was pretty large as is.

shared/address_explorer.py
```diff
@@ -197,7 +187,15 @@ def generate_address_csv(path, addr_fmt, n):
         return
 
     path, addr_fmt = picked
+    chain = chains.current_chain()
+    def get_address_at_idx(idx):
+        with stash.SensitiveValues() as sv:
```
peter-conalgo: This change will impact performance heavily. `stash.SensitiveValues()` context manager takes a long time to start up because it talks via bootrom to the SE. Moving it outside the loop like this will be terrible (on real hardware).

shared/multisig.py
```diff
@@ -889,10 +889,94 @@ def update_contents(self):
...
+        AF_P2SH: ([HARD(45), is_change, idx], "m/45'/{change}/{idx}".format(change=is_change, idx=idx)),
+        AF_P2WSH_P2SH: (
+            [HARD(48), HARD(coin), HARD(0), HARD(1), is_change, idx],
+            "m/48'/{coin}'/0'/1'/{change}/{idx}".format(coin=coin, change=is_change, idx=idx)
+        ),
+        AF_P2WSH: (
+            [HARD(48), HARD(coin), HARD(0), HARD(2), is_change, idx],
+            "m/48'/{coin}'/0'/2'/{change}/{idx}".format(coin=coin, change=is_change, idx=idx)
+        )
+    }
```
peter-conalgo: Two thirds of this work is thrown out after a few more lines. Little tiny micro had to calculate it, and store it all first though.

shared/multisig.py
```diff
@@ -889,10 +889,94 @@ def update_contents(self):
...
+    mm = [80 + M] if M <= 16 else [1, M]
+    nn = [80 + N] if N <= 16 else [1, N]
+    rv = bytes(mm)
+    for pk in pubkeys:
+        rv += bytes([len(pk)]) + pk
+    rv += bytes(nn + [0xAE])
+
+    return rv, pubkeys
+
+def ms_path_mapper(idx, is_change, coin, addr_fmt):
```
peter-conalgo: Please don't pass around 'coin' ... instead, just lookup the current chain when needed. Affects this function and a few others nearby.

shared/multisig.py
```diff
@@ -889,10 +889,94 @@ def update_contents(self):
...
+        raise ValueError("Unsupported address format: %s " % addr_fmt)
+    return d[addr_fmt]
+
+def make_ms_address(M, keys, idx, is_change, coin, addr_fmt):
+    from chains import BitcoinMain, BitcoinTestnet
+    # Validations
+    assert M <= len(keys), 'M > N'
+    assert idx >= 0, 'Invalid idx'
+    assert is_change in {0,1}, 'Invalid change value'
+    assert coin in {0, 1}, 'Invalid coin type'
```
peter-conalgo: Remove and use current chain data instead.

peter-conalgo:
Sorry I pressed "start review" by mistake, and no clear way to un-do it... meant to just comment on some line numbers.

shared/address_explorer.py
hodlwave: Makes sense. I could try refactoring it to `get_addresses(from, to)` so that all the addresses for a given screen / text file backup will be calculated in the same `stash.SensitiveValues()` context. Sound good?

shared/multisig.py
hodlwave: Noted

hodlwave:
#### New updates

Implemented fixes based on @peter-conalgo PR review. These include:
- Refactor get_address_at_idx to get_addresses for performance
- Avoid passing coin as argument to methods
- Avoid calculating extra data that is never used

peter-conalgo:
Great changes, thanks.

I've looked more closely now, and here are my concerns:

1) I don't like `ms_path_mapper()` because it presumes "purpose 48" being used, and we don't assume that elsewhere. We don't need to reconstruct paths because we know the shared common prefix path based on the wallet (`common_prefix`). .. we can just take that and add `{change}/{idx}`. I suggest deleting `ms_path_mapper()` .. sorry.
2) `make_ms_address()` should be a member function on `MultisigWallet`, and should probably take a "suffix path" to be appended onto the existing common prefix path (`self.common_prefix`). For this explorer application, we'd provide string: '0/0' thru '0/10'... The `keys` argument should stay as an optimization.
3) IIRC some multisig schemes use:   `...common prefix.../{index of co-signer}/{change}/{idx}`, so that there is no danger of collision between the co-signers when they construct addresses. We should support that. I don't see how we can though. 

I am very scared of providing an address the UTXO-watching wallet isn't expecting. With multisig, we might even make an address that we cannot recover funds from, because we need additional signatures from another system that is not as flexible as we are with derivation paths.

hodlwave:
Good points. I think they actually speak to a bigger issue I hadn't considered. Since the MultisigWallet only stores up to the `common_prefix`, it doesn't know anything about the full derivation used until instructed by the PSBT during transaction signing.

So the open question is what derivations do you want to choose to show in an Address Explorer if any? Allowing the user to explore more derivation paths increases complexity. But reducing this ability or eliminating it entirely prevents users from verifying receive addresses on device.

What do you think?

destrys:
I don't believe Electrum allows you to specify custom suffixes, it's just `/0/*` and `/1/*`. That could be a good place to start, maybe with an explicit warning that the address explorer is only displaying electrum-compatible addresses?

We're excited about this feature...

peter-conalgo:
I just had an idea: what if we only show part of the address. Maybe the left half of it... then the feature is still useful for verification purposes, but customers cannot deposit funds to an address they cannot sign for. The QR stuff would be disabled as well.

destrys:
I don't see the benefit of displaying a partial address. (Especially since users should confirm every character in an address before interacting with it). The left side doesn't include the checksum so I wouldn't use it for verification, need every character for verification.

I don't fully understand the worry about displaying addresses derived from the multisig config file. If the device is confident that it can sign for at least one of the xpubs (the one with the same xfp as the device), then displaying the address on device is only claiming that *this* device can produce signatures for the shown address, it's not claiming anything about the other keys, since it can't. It would be even better if the device displayed the multisig shape (2 of 3) as well since there is the worry about an attacker adding keys to the script (making it 3 of 5 where they control 3 instead of 2of3 where a collaborator holds 1)

But during multisig config file loading, does the device validate that at least one xpub included in the config file is actually signable by the device? If not, then I understand the worry.

peter-conalgo:
Even better idea: We'll show the whole address, except for the last four characters (ie. the checksum). Then we have a full-verifiable value but still not be showing a directly useable address. Since the checksum is implied by the other digits, there is no way for an attacker on the desktop to construct a precisely-matching address. (Not sure what that will look like for bech32 based addresses, but I think the same idea applies.)

> I don't fully understand the worry about displaying addresses derived from the multisig config file. If the device is confident that it can sign for at least one of the xpubs...

Signing for one leg is assured, but we have no idea what the other legs are capable of. They could have arbitrary policies about derivation paths that don't follow standards. The same situation doesn't exist for single-signers because we always know that we'd be able to construct a PSBT to recovery misdirected funds, using just the Coldcard and no external wallets. 

BTW, we do get support tickets for this sort of accident, and it's not too hard to get the funds "back" by hand-constructing a suitable PSBT file for the unique situation. Our tool for this is available here: [PSBT recovery](https://github.com/Coldcard/psbt_recovery)

hodlwave:
I think there are a few considerations for maximizing security when receiving bitcoins to a multisig address.

1) The user should register the multisig quorum with each cosigner. For receiving, the multisig conig might be fully specified by an external output descriptor like `wsh(sortedmulti(M,[aabbccdd/path1]xpub1/0/*,[bbccddee/path2]xpub2/0/*,[ccddeeff/path3]xpub3/0/*))` and a range of indices (e.g. `[0, 1000]`).
2) Each time the user wants to deposit bitcoins to a new address, he should have at least M offline signers display the address.
3) If M devices agree about the address and assuming they aren't _all_ compromised, he should be secure in receiving the bitcoins to that address.

If these conditions aren't met, the user is basically just hoping that he's sending bitcoins to an address he controls. With that in mind, in my opinion the best multisig UX would try to meet these conditions comprehensively.

@peter-conalgo, I understand your concern about this issue because users could be at risk of losing funds due to deficiencies of other wallets; unfortunately I think this is just a reality of multisig, and just showing the address minus checksum isn't a robust solution to this problem.

lucasmoten:
I second this 
> They could have arbitrary policies about derivation paths that don't follow standards.
There doesn't seem to be formalized standards (yet), but it seems 44, 45, 48, 49 purpose are common though some with and without hardening

I think its fair to support defaults for Coldcard multisig, but it would be nice to permit the user to specify derivation path prefix for edge cases.

nvk:
Done via #61 & #60

PR #32, opened 2019-11-18
Add Bitcoin Core descriptor export

Sjors:
This PR may have broken some tests; I can clean that up later, after initial feedback.

Bitcoin Core 0.18 a later added support for [output descriptors](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md) and watch-only wallets. This makes it much easier import keys from Coldcard. In fact, it's what [HWI](https://github.com/bitcoin-core/HWI) uses internally.

This PR allows air gapped import of public keys, by adding a new menu item: 
Schermafbeelding 2019-11-18 om 13 49 03

This produces a text file with a command, that can be copy-pasted to import public keys into Bitcoin Core. This is the same command HWI produces in the [setup step](https://github.com/bitcoin-core/HWI/blob/master/docs/bitcoin-core-usage.md#setup), but doesn't require an USB connection.

```
# Bitcoin Core Wallet Import File

https://github.com/Coldcard/firmware/blob/master/docs/bitcoin-core-usage.md

## For wallet with master key fingerprint: 0f056943

Wallet operates on blockchain: Bitcoin Testnet

## IMPORTANT WARNING

Do **not** deposit to any address in this file unless you have a working
wallet system that is ready to handle the funds at that address!

## Bitcoin Core RPC

The following command can be entered after opening Window -> Console in Bitcoin Core,
or using bitcoin-cli:

importmulti '[{"range": [0, 1000], "timestamp": "now", "keypool": true, "watchonly": true, "desc": "wpkh([0f056943/84h/1h/0h]tpubDC7jGaaSE66Pn4dgtbAAstde4bCyhSUs4r3P8WhMVvPByvcRrzrwqSvpF9Ghx83Z1LfVugGRrSBko5UEKELCz9HoMv5qKmGq3fqnnbS5E9r/0/*)#erexmnep", "internal": false}, {"range": [0, 1000], "timestamp": "now", "keypool": true, "watchonly": true, "desc": "wpkh([0f056943/84h/1h/0h]tpubDC7jGaaSE66Pn4dgtbAAstde4bCyhSUs4r3P8WhMVvPByvcRrzrwqSvpF9Ghx83Z1LfVugGRrSBko5UEKELCz9HoMv5qKmGq3fqnnbS5E9r/1/*)#ghu8xxfe", "internal": true}]'

```

This RP also adds more detailed Bitcoin Core instructions. It removes previous instructions that imo are too complicated and error prone. However I can add them back if needed, e.g. under an "older versions" heading.

Like Wasabi this only supports bech32 addresses. Mixing and matching P2SH and bech32 addresses in Bitcoin Core currently breaks compatiliity with BIP59 / BIP84. Future [descriptor wallets](https://github.com/bitcoin/bitcoin/pull/16528) will be able to mix and match safely, if anyone still uses P2SH wrapped SegWit by then.

To add PSBT loading and saving support to Bitcoin Core GUI, please review: https://github.com/bitcoin/bitcoin/pull/17509. Until then, PSBT support is command-line only and uses the JSON instead of binary format. I added a link to HWI documentation for how to sign transactions.

### DIFF
--- a/shared/backups.py
+++ b/shared/backups.py
@@ -527,21 +527,13 @@ be needed for different systems.
             yield fp.getvalue()
             del fp
 
-async def make_summary_file(fname_pattern='public.txt'):
-    # record **public** values and helpful data into a text file
+# total_parts does need not be precise
+async def write_text_file(fname_pattern, body, title, total_parts=72):
     from main import dis, pa, settings
     from files import CardSlot, CardMissingError
     from actions import needs_microsd
 
-    dis.fullscreen('Generating...')
-
-    # generator function:
-    body = generate_public_contents()
-
-    total_parts = 72        # need not be precise
-
     # choose a filename
-        
     try:
         with CardSlot() as card:
             fname, nice = card.pick_filename(fname_pattern)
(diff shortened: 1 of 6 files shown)


shared/backups.py
```diff
@@ -559,9 +551,92 @@ def generate_public_contents():
...
+    return list(map(lambda internal: {
+        'desc': AddChecksum("wpkh([{fingerprint}/84h/{coin_type}h/{account}h]{xpub}/{change}/*)".format(
+            fingerprint=txt_xfp,
+            coin_type=chain.b44_cointype,
+            account=0,
+            xpub=xpub,
+            change=1 if internal else 0
+        )),
+        'range': [0, 1000],
+        'timestamp': 'now',
```
Sjors: Does Colcard have a clock and know when the seed was generated? If so, that could be added here.

peter-conalgo: Sadly Coldcard has no idea of time nor block height.

peter-conalgo:
This looks great and is obviously useful. I will review and get feedback to you asap.

peter-conalgo:
@Sjors Please review my test case  (`testing/test_export.py`) and if you can tighten it, please submit a PR for that. Thanks again for adding this feature.

Sjors:
Thanks! The test looks OK, but maybe you want to add a test that interacts with bitcoind, and checks that `getnewaddress` returns the correct first address.

Another useful feature I think would be to print the first couple of addresses in the txt file, so users can more easily check that they didn't mess up the import.

peter-conalgo:
Yes, on both counts.

Sjors:
Also, if you want to add support for airgapped transaction signing, you could add support for reading a plain text file with the base64 encoded PSBT.

And for the signed result, you could even add the required command, I believe the GUI console lets you nest stuff: sendrawtransaction(finalizepsbt(PSBT))
But I haven't tried

peter-conalgo:
Coldcard supports binary PSBT files coming from MicroSD already. If the result is a finalized transaction (as in typical single-signer case), then the hex-encoded transaction is also saved to MicroSD card.

peter-conalgo:
I can import the keys into bitcoind, but I don't have a good way to test they got in there.
- `getnewaddress` ... theres lots of crap in my bitcoind, and I don't expect it to return same thing everytime; I don't see a way to control what pool of addresses it draws from.
- `getaddressinfo` ... did not return anything useful:
```python
{'address': 'tb1qxvzlcjr2djne6n60q0ytzet49mcqumpdqmluzm',
 'ischange': False,
 'ismine': False,
 'isscript': False,
 'iswatchonly': False,
 'iswitness': True,
 'labels': [],
 'scriptPubKey': '00143305fc486a6ca79d4f4f03c8b165752ef00e6c2d',
 'solvable': False,
 'witness_program': '3305fc486a6ca79d4f4f03c8b165752ef00e6c2d',
 'witness_version': 0}
```
In particular:
- `label` isn't set, but perhaps that's because no label was set first time I imported these keys?
- `ismine` and/or `iswatchonly` aren't true?

I am using 0.18.1

Sjors:
You should create a fresh bitcoind wallet for this, and it needs to be watch-only (see `createwallet` wallet command in the docs that I added). In order to select that wallet from RPC, you need to add `-rpcwallet=...` to the command. You should also `rm -f` it before the start of the test run.

> Coldcard supports binary PSBT files coming from MicroSD already.

But Bitcoin Core doesn't. So users can only put a (unsigned) base58 encoded PSBT on the SD card.

> If the result is a finalized transaction (as in typical single-signer case), then the hex-encoded transaction is also saved to MicroSD card.

Ok, so then the user can call `sendrawtransaction` on that.

peter-conalgo:
- I've created a text fixture to make a temp wallet, and added that to the test case
- support for Base64 PSBT files will come in a future version of Coldcard

Sjors:
a2329cf8e6591ca3a276ebc700f81ab71b7eeeaf looks good to me. If you use #32 in the commit description, Github will automatically add a comment to this thread linking to it.

You could use this to make the tests works on Windows and Linux: https://github.com/cryptoadvance/specter-desktop/blob/aa0513f5521f672bc24e6e615d07dd5bd8c6a103/src/rpc.py#L8-L16

PR #36, opened 2020-01-21
Support firmware limiting itself to specific hardware revs

peter-conalgo:
- add a new field to firmware header, which indicates which hardware versions it supports (`hw_compat`)
- designed to break the signature on the firmware image, so that fielded bootroms will not install those new versions
- it makes this version (3.0.7) a required upgrade, because no future version of the firmware will load without the patches it applies during firmware upgrade process
- difficult change, still somewhat WIP


peter-conalgo:
merged into branch `hsm` where it will join master.

PR #42, opened 2020-05-05
Add guide to fix "Skip PIN" bug on Linux

RiccardoMasutti:


peter-conalgo:
- might be better as `if 0:` rather than comments (because multiline)
- any chance of auto-detect?

RiccardoMasutti:
@nvk ?

peter-conalgo:
Helpful! Thanks.

PR #43, opened 2020-05-10
Multisig descriptor wallet

hodlwave:
This gives users the ability to generate a multisig wallet by importing a text file containing an [output descriptor](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md). An advantage of using output descriptors is that they provide an unambiguous description of a wallet's addresses.

The following is an example of a valid text file input:
```
Name: wsh descriptor 1/2
Descriptor: wsh(sortedmulti(1,[598da155/48'/1'/0'/2']tpubDFcMAKmwXTqkk19fKuaxFc1P35NYg2efhqsCyZa1h2g5PmYjvzpiM1v2yrTSBxY54v6Zawp96BHQWSC7ZxXbFXNxRY93MnmzSc2Wc9DTRYC,[0f056943/48'/1'/0'/2']tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP))
```

The user must fully specify the key origin to pass validations. We assume the `/change/idx` after the account xpubs to avoid requiring separate descriptors for the `external` and `internal` paths. This is consistent with how [specter-desktop](https://github.com/cryptoadvance/specter-desktop/) exports wallet descriptors, and seems like a reasonable standard to conform to. Any other keys in the text file besides `Name` and `Descriptor` will be ignored. 

In accordance with the Coldcard's current functionality, this PR only supports the following three descriptors:
```
sh(sortedmulti(...))
```
```
sh(wsh(sortedmulti(...)))
```
```
wsh(sortedmulti(...))
``` 

A concern raised in https://github.com/Coldcard/firmware/pull/25 was that the Coldcard would generate an address that the watch-only wallet isn't expecting. Since the output descriptor solves this problem unambiguously, we enable the multisig address explorer for a wallet imported with a descriptor. 

In the multisig address explorer, addresses are prefixed with the full derivation path from the Coldcard's root xprv to the index of the associated address. Example:
```
Addresses 0..9
m/48'/1'/0'/2'/0/0 => tb1qym0lechjavp3fmfuy5yr3kfqrsyfvegd0jqcfwpvc0anj9xq5arstm8dgk
```
Since other cosigner's in the multisig quorum don't have to share the same hardened derivation path, it might make more sense to show the user something like this:
```
Addresses 0..9
.../0/0 => tb1qym0lechjavp3fmfuy5yr3kfqrsyfvegd0jqcfwpvc0anj9xq5arstm8dgk
```

A sensible follow-up to this PR would be the addition of a UX flow for exporting the Coldcard's key origin at a desired hardened derivation path to a text file. This would make the Coldcard fully compatible with [specter-desktop](https://github.com/cryptoadvance/specter-desktop/) in airgapped mode.


stepansnigirev:
We drop `/0/*` and `/1/*` part in the receiving and change descriptors because of the QR code size limit (to avoid duplication of almost the same descriptor), but it is not standard and not compatible with how Bitcoin Core imports descriptors.
As ColdCard doesn't have a space problem, I suggest importing internal and receiving descriptors separately and making them compatible with Core.
So instead of a single descriptor you could have two separate descriptors - for receiving and change addresses. What do you think?
From our side we will be happy to replace hacks we used "save coldcard file" to normal core-like descriptors.

hodlwave:
It's a fair call-out about using separate receiving and change descriptors. It's better to keep consistent with Bitcoin Core. Also that reminds me that this PR doesn't include any changes to the signing flow to guarantee that outputs being signed are consistent with the implicit `/0/*` chain while change outputs only go back to the implicit `/1/*` chain. If I'm not mistaken, having this level of explicitness would also solve the issue of ransom attacks to hard to find change paths.

Without strict validations in place, letting the user specify separate descriptors might give them a little too much rope to hang themselves with... Here are some things we might reasonably want to protect them from doing:
* Use different descriptor types `sh(sortedmulti(...))` vs `sh(wsh(sortedmulti(...)))` vs `wsh(sortedmulti(...))` for each descriptor
* Pubkey reuse or address reuse
* Use different hardened derivation paths for a single cosigner's receiving vs change
```
descriptor_external: wsh(sortedmulti(1,[598da155/48'/1'/0'/2']tpubDFcMAKmwXTqkk19fKuaxFc1P35NYg2efhqsCyZa1h2g5PmYjvzpiM1v2yrTSBxY54v6Zawp96BHQWSC7ZxXbFXNxRY93MnmzSc2Wc9DTRYC/0/*,[0f056943/48'/1'/0'/2']tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP/0/*))
descriptor_internal: wsh(sortedmulti(1,[598da155/100']tpubDFcMAKmwXTqkk19fKuaxFc1P35NYg2efhqsCyZa1h2g5PmYjvzpiM1v2yrTSBxY54v6Zawp96BHQWSC7ZxXbFXNxRY93MnmzSc2Wc9DTRYC/1/*,[0f056943/100']tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP/1/*))
```
* Use different unhardened paths from the account xpub to the pubkey within the same descriptor and/or across the two descriptors
```
descriptor_external: wsh(sortedmulti(1,[598da155/48'/1'/0'/2']tpubDFcMAKmwXTqkk19fKuaxFc1P35NYg2efhqsCyZa1h2g5PmYjvzpiM1v2yrTSBxY54v6Zawp96BHQWSC7ZxXbFXNxRY93MnmzSc2Wc9DTRYC/0/*,[0f056943/48'/1'/0'/2']tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP/0/0/0/*))
descriptor_internal: wsh(sortedmulti(1,[598da155/48'/1'/0'/2']tpubDFcMAKmwXTqkk19fKuaxFc1P35NYg2efhqsCyZa1h2g5PmYjvzpiM1v2yrTSBxY54v6Zawp96BHQWSC7ZxXbFXNxRY93MnmzSc2Wc9DTRYC/1/1/1/*,[0f056943/48'/1'/0'/2']tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP/1/1/*))
```
* General use of non-standard unhardened derivations paths to pubkeys

These schemes could technically be permitted but I assume for 99% of users, it's probably safer just to not let them do these kinds of things.

Curious what @nvk @peter-conalgo think of the PR in general and also the idea of separating the receive and change descriptors.

stepansnigirev:
I agree with you. Maybe this style of descriptors is better from a security standpoint.

peter-conalgo:
I am basically ok with this PR, but my concerns:
- [ ] we are out of flash space and so all new additions need to be tight
- [ ] replace `if not cond: raise AssertionError(..)` with regular `assert cond, 'err msg'`
- [ ] shorten all error messages to absolute minimum, and/or remove their text (still have the assert tho, it will show as file:line which is all thats needed to debug in the field) ... and fix all test cases to match same (or maybe dont put text of error into test case, IDK).
- [ ] I'd prefer `DESCRIPTOR_PATTERNS` was achieved with a regex
- [ ] did you implement the "export multisig as descriptors" feature? Seems like a natural

hodlwave:
Cool, I'll make those changes.

For the "export multisig as descriptors" feature, did you mean...
a) exporting a key origin at a specified derivation path as a first step to generating a multisig wallet in specter-desktop or Electrum
`key-origin.txt`
```
[0f056943/48'/1'/0'/2']tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP
```
b) exporting the whole file used to generate the descriptor multisig wallet
`multisig-export.txt`
```
Name: wsh descriptor 1/2
Descriptor: wsh(sortedmulti(1,[598da155/48'/1'/0'/2']tpubDFcMAKmwXTqkk19fKuaxFc1P35NYg2efhqsCyZa1h2g5PmYjvzpiM1v2yrTSBxY54v6Zawp96BHQWSC7ZxXbFXNxRY93MnmzSc2Wc9DTRYC,[0f056943/48'/1'/0'/2']tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP))
```
c) all of the above

hodlwave:
I forgot that I initially implemented `DESCRIPTOR_PATTERNS` with regexes, but I get a `maximum recursion depth exceeded error` when the descriptor gets sufficiently large.

peter-conalgo:
1. recursion depth sounds fixable if you can accommodate the limitations of `ure` module
2. for exporting multisig descriptors ... I'm not sure what's most useful, but I had core in mind. perhaps as  a sequence of CLI commands?

hodlwave:
1. From what I've [read](https://github.com/micropython/micropython/issues/3574), the max recursion depth reached is a consequence of running `ure` in a low-memory environment. If the work around is [setting the recursion limit](https://stackoverflow.com/a/3323013/12956812) to avoid a stack overflow, it seems safer to just match the exact prefix and suffix. Thoughts?

2. I followed @achow101's [instructions](https://gist.github.com/achow101/a9cf757d45df56753fae9d65db4d6e1d) to import the descriptor wallet for use with a hardware wallet; however, unfortunately it doesn't look like the user can generate new receive addresses or create unsigned PSBTs with multisig descriptors in `bitcoin-qt` yet. Not sure why as it works with the `wpkh` example from the gist. Maybe sideline this feature for now until the GUI has full watch-only functionality?

I used the following commands to import the descriptor to Bitcoin Core:

// create wallet
```
bitcoin-cli -testnet createwallet "blah" true
```

// import descriptors
```
bitcoin-cli -testnet -rpcwallet="blah" importmulti '[{"desc": "wsh(sortedmulti(1,[598da155/48h/1h/0h/2h]tpubDFcMAKmwXTqkk19fKuaxFc1P35NYg2efhqsCyZa1h2g5PmYjvzpiM1v2yrTSBxY54v6Zawp96BHQWSC7ZxXbFXNxRY93MnmzSc2Wc9DTRYC/0/*,[0f056943/48h/1h/0h/2h]tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP/0/*))#rah27k5g", "internal": false, "range": [0, 100], "timestamp": "now", "keypool": true}]' 
```

```
bitcoin-cli -testnet -rpcwallet="blah" importmulti '[{"internal": true, "timestamp": "now", "desc": "wsh(sortedmulti(1,[598da155/48h/1h/0h/2h]tpubDFcMAKmwXTqkk19fKuaxFc1P35NYg2efhqsCyZa1h2g5PmYjvzpiM1v2yrTSBxY54v6Zawp96BHQWSC7ZxXbFXNxRY93MnmzSc2Wc9DTRYC/1/*,[0f056943/48h/1h/0h/2h]tpubDF2rnouQaaYrXF4noGTv6rQYmx87cQ4GrUdhpvXkhtChwQPbdGTi8GA88NUaSrwZBwNsTkC9bFkkC8vDyGBVVAQTZ2AS6gs68RQXtXcCvkP/1/*))#6wyws9pa", "keypool": true, "range": [0, 100]}]'
```

achow101:
>     2\. I followed @achow101's [instructions](https://gist.github.com/achow101/a9cf757d45df56753fae9d65db4d6e1d) to import the descriptor wallet for use with a hardware wallet; however, unfortunately it doesn't look like the user can generate new receive addresses or create unsigned PSBTs with multisig descriptors in `bitcoin-qt` yet. Not sure why as it works with the `wpkh` example from the gist. Maybe sideline this feature for now until the GUI has full watch-only functionality?

This functionality should work with descriptor wallets with the `importdescriptors` command. Legacy wallets have limitations that prevent non-single key things from working in this way.

hodlwave:
Oh perfect; thanks, Andrew.

Looks like we'll have to wait until Bitcoin Core 0.21 to use this rpc command.

PR #44, opened 2020-05-15
fixed typo in comments

zachherbert:
I believe fees over 5% are now warned, not 1%. Please confirm this is a typo!


peter-conalgo:
Correct. Check `"shared/psbt.py" line 1125` where 5% threshold is enforced.

PR #46, opened 2020-06-10
Resist recently reported weakness of BIP-143

peter-conalgo:
- Enhancement: Detect, report and block the recently reported type of attack
  against BIP-143 (replay of segwit inputs) with an error message. No changes
  needed to your input PBST files. Will show errors similar to:
  "Input#0: Expected 15 but PSBT claims 5.00001 BTC"
- Enhancement: When the Coldcard is finalizing the transaction, we show the TXID (hex
  transaction ID) of the transaction on the screen.


peter-conalgo:
- [ ] test on real hardware

peter-conalgo:
- [x] store some UTXO in ram during operation
- [ ] bug: flash of main menu before TXID is shown

PR #50, opened 2020-10-15
Add guidance if PSBT file is too large

lucasmoten:


peter-conalgo:
Thanks for this.

Feedback:
- I would perfer `is_psbt()` to count oversized files that it tastes. Then we wouldn't need to run twice over all the files.
- The real problem is usually too many inputs, not file size exactly. A minimal PSBT parser that could count number of ins and outs would be better. 
- The file size is a hard limit, so that can fail, but a better fix would warn about large transactions (before failing) and warn the user about splitting it up and such, but still allow them to proceed.... because there are many factors, and it "might work"

lucasmoten:
Thanks Peter,

I like your recommendations and will take another crack at this over the coming days.

peter-conalgo:
Feel free to reopen as needed. Just tidying up here.

PR #51, opened 2020-10-15
Restore MIT licence

switck:
want to use this file in something else, and you just copied it anyway


peter-conalgo:
Why not just take from < https://github.com/micropython/micropython-lib/blob/master/hmac/hmac.py> then?

switck:
there are some diffs

peter-conalgo:
Well I hope you're writing something good! Merging...

PR #55, opened 2020-11-23
Adds p2sh/p2wsh as valid AddressType

humanumbrella:
This is inverted form most places I've seen and I ran into it while integrating Coldcard into [Caravan](https://github.com/unchained-capital/caravan). It's totally fine to define it this way, but it would be nice to seamlessly support the other order as well.


humanumbrella:
I did not test this change, more just offering a suggestion.

peter-conalgo:
You are right, the convention is `p2sh-p2wsh` ... I've made the change, but for reasons, didn't use this PR.

Thanks for reporting this! I will honour you in the release notes.

PR #57, opened 2020-12-21
Add warnings for weak input to rolls.py

lucasmoten:


peter-conalgo:
Thanks! Useful warnings, just change:
- `emptyhhex` => `empty`
-  minimum rolls from 100 to 99 ... the value Coldcard uses as "good enough"

PR #63, opened 2021-02-15
Increase BOOT_LOCKOUT_TIME to escape HSM.

lucasmoten:
Justification
30 seconds is insufficient when using features that consume time (Scrambled Keypad, setting a Nickname, and especially Login Timeout)

Consider this scenario with a nickname and scrambled pin entry
The following all needs to be accomplished within the first 30 seconds (not physically possible)
1. wait for green light. click yes
2. wait for custom label. click yes
3. enter pin prefix on scrambled pad. click yes
4. wait for anti-phishing words. click yes
5. enter pin suffix on scrambled pad. click yes
6. wait for HSM policy prompt. click yes
7. then enter 6 digit pin. click yes

### DIFF
--- a/shared/hsm.py
+++ b/shared/hsm.py
@@ -30,7 +30,7 @@ ABSOLUTE_MAX_REFUSALS = const(100)
 
 # you have this many seconds after boot to escape HSM
 # mode, if you enable the boot_to_hsm feature
-BOOT_LOCKOUT_TIME = const(30)
+BOOT_LOCKOUT_TIME = const(45)
 
 def hsm_policy_available():
     # Is there an HSM policy ready to go? Offer the menu item then.


peter-conalgo:
Why did you close this? I thought it was worth discussing... and I was going to advocate for a full minute.

lucasmoten:
I didn't yet have a way to adequately test it on the unit I'm playing with so it seemed premature to pose this as a change.
I still think the timeout should be longer, and 1 minute would likely suffice for cases other than if Login Countdown is set.

A compromise on Login Countdown could be to add that time to the check in hsm.py local_pin_entered

https://github.com/Coldcard/firmware/blob/c1d78d12528d7c4b0f12c3a4ea6c18453d424f5e/shared/hsm.py#L680

e.g.
```py
if self.boot_to_hsm and self.uptime < BOOT_LOCKOUT_TIME + settings.get('lgto', 0)
```

peter-conalgo:
period has been doubled to 1minute, on v4.0.0b3 version
going to ignore interaction with login countdown mode for now, but will make a note to fix that later

peter-conalgo:
- actually, does not interact with login countdown; the "uptime" of the HSM doesn't start counting until it's activated which occurs shortly after second PIN entry, not at start of countdown.

PR #67, opened 2021-03-10
Firmware V4

nvk:


lucasmoten:
I've loaded this on an Mk3 to check out.  I don't know if its something wonky in this particular unit, or if its the new modules, but I'm getting very sporadic responsiveness for menu navigation.

peter-conalgo:
Known issue, working on it.

lucasmoten:
Is it a dramatically reduced clockspeed ? Button presses require me to hold a button down for about 5 to 10 seconds, then release, and it registers.  Loading screens take .... minutes/hours, hopefully not days

shared/queues.py
```diff
@@ -26,7 +25,7 @@ class Queue:
     with qsize(), since your single-threaded uasyncio application won't be
     interrupted between calling qsize() and doing an operation on the Queue.
     """
-    _attempt_delay = 0.01
+    _attempt_delay = 10         # milliseconds
```
lucasmoten, line 28: Curious about this change in delay being 1000x its prior value.

peter-conalgo: it now uses `sleep_ms()` rather than `sleep()` so 0.010 seconds == 10 ms.

jmassu:
I also installed, and device takes ages for everything.
I'm trying to re install latest version, and I hope it ends this year. Will it?
Please help.

lucasmoten:
> I also installed, and device takes ages for everything.
> I'm trying to re install latest version, and I hope it ends this year. Will it?
> Please help.

On my unit it takes about 4 hours to progress through firmware update. You should see your progress bar moving very slowly.

lucasmoten:
@jmassu I just realized that simple downgrading to older dfu won't work here. Need a firmware update with newer timestamp.

peter-conalgo:
This performance issue has been fixed in `2021-03-15T1304-v4.0.0b3`, available now. To workaround the issue (during upgrade for example), you just need to connect the Coldcard to a USB host (even if not being used). The issue occured during "air gapped" mode, where only power is coming in USB.

peter-conalgo:
I think this will be the last beta version. Please consider 4.0.0b4 a release candidate.

PR #69, opened 2021-04-14
repro: fix dns redirect issue in dockerfile.build

RandyMcMillan:
The dns redirect for the edge/testing seems unstable/broken.
This causes and error when building the repro.


nvk:
Thx!

PR #74, opened 2021-06-08
changed %d to float to mitigate rounding issues in Shannon entropy

adlererik:
Changed the way Shannon entropy was being presented.

Example:
echo 123 |python3 rolls.py was displaying 7 bits of entropy when the actual entropy was much closer to 8 (7.755...)


adlererik (commented):
added a space \n before word list.

doc-hex:
f-strings are not supported before 3.6 so this change is reducing the number of places the code can run.

PR #85, opened 2021-11-08
Add workflow for repro build on github actions

JS1010111:
Hi guys,

I made this workflow for the repro build of the Coldcard firmware that is automatically triggered when a commit or pull request is merged into master.

It runs on Github runners with Ubuntu but it's easy to add other OSes if you want or even run on Docker with your dockerfile (though a bit slower and imho it's better to avoid another layer).

[Github Actions](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) is free for public repos, so you get a basic CI without any cost. It even gets you a badge if you want to stamp it on your readme like this:

[CCFW Reproducible Build](https://github.com/JS1010111/firmware/actions/workflows/ccfw-repro-build.yml)

Unfortunately the reproducible build from master is failing because you have commited new code that is not in the latest released firmware v4.1.3, so the diff fails.

It's better to keep your master branch reflecting your production state and use branches for everything else if the reproducible build comes from the master. It's also possible to trigger the CI from tags or release/\<version\> branches if you will.

Please feel free to modify this PR or ask me for any changes / enhancements. I hope it'll be useful!

### DIFF
--- a/.github/workflows/ccfw-repro-build.yml
+++ b/.github/workflows/ccfw-repro-build.yml
@@ -1,0 +1,44 @@
+name: CCFW Reproducible Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Git Checkout
+      uses: actions/checkout@v2
+      with:
+        submodules: true
+        
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: "3.10"
+        
+    - name: Setup Environment
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y gcc-arm-none-eabi
+        python -m pip install virtualenv
+        virtualenv -p python3 ENV
+        source ENV/bin/activate
+        
+    - name: Build Firmware
+      run: |
+        cd $GITHUB_WORKSPACE/stm32
+        VERSION_STRING=$(awk -F ' = ' '/^VERSION_STRING = /{ print $2 }' Makefile)
+        chmod +x repro-build.sh
+        ./repro-build.sh $VERSION_STRING $GITHUB_WORKSPACE $GITHUB_WORKSPACE/built
+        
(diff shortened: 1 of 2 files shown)


JS1010111:
Almost forgot... you can take a look at a run logs on my fork https://github.com/JS1010111/firmware/actions

doc-hex:
Can we make this happen only when a release is tagged? We don't generally expect the tip of master to be reproducible.

JS1010111:
I'm sure it's possible but it seems that github doesn't make it trivial. I'll make it work on a test repo and get back to you soon.

JS1010111:
This works:

```yaml
on:
  push:
    tags:
      - '**'
      - '!**bootstrap*'
```

So it silently checks these rules everytime you push something. In this case, it's defined to only be triggered when you push tags.

It uses glob patterns so you can include/exclude tag names (order matters, we include everything then excludes what contains `bootstrap` at any level).

What may be confusing is that Github Actions reads the workflow yaml from the branch you have created the tag.

That is if you tag and push from a branch without the workflow file `.github/workflows/ccfw-repro-build.yml` it will not be triggered even if you have it on master and can see the workflow on the Actions tab.

If you are only tagging from master there's nothing to worry about after merging the PR.

.github/workflows/ccfw-repro-build.yml
```diff
@@ -0,0 +1,44 @@
...
+      - '**'
+      - '!**bootstrap*'
+  
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Git Checkout
+      uses: actions/checkout@v2
```
scgbckbone, line 15: can be bumped to `actions/checkout@v3`

.github/workflows/ccfw-repro-build.yml
```diff
@@ -0,0 +1,44 @@
...
+      with:
+        submodules: true
+        
+    - name: Setup Python
+      uses: actions/setup-python@v2
```
scgbckbone, line 20: can be bumped to `actions/setup-python@v3`

PR #88, opened 2022-01-25
Update COPYING-CC

MarnixCroes:
Update Copyright 2020 to Copyright 2022


doc-hex:
IANAL, but I think the first date it was published is supposed to be on the (C) notice. Maybe a range, IDK.

PR #93, opened 2022-04-18
doc: bitcoin core usage improvement

mjdietzx:
If the Disable Private Keys option is not checked when creating
the watch-only wallet for the Coldcard, the `importdescriptor`
command that the user copy-pastes in Step 2 will fail with an
error "Cannot import descriptor without private keys to a wallet
with private keys enabled"

This commit instructs the user to check that option so everything
in this doc works as expected, and adds some explanation in a note.


doc-hex:
Thanks for this @mjdietzx !

PR #96, opened 2022-05-03
Fix unix/variant/ffilib.py to detect linux systems correctly

achow101:
It appears that sys.platform on linux systems report as "coldcard-unix" so ffilib.py needs to search for the string "unix" in order for the simulator to work on linux.

### DIFF
--- a/unix/variant/ffilib.py
+++ b/unix/variant/ffilib.py
@@ -14,7 +14,7 @@ def open(name, maxver=10, extra=()):
     except KeyError:
         pass
     def libs():
-        if sys.platform == "linux":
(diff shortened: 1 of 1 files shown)


doc-hex:
thoughts @scgbckbone ?

scgbckbone:
known issue already handled [here](https://github.com/Coldcard/firmware/pull/91), we were waiting with merge for ngu to accept our PR (which happened yesterday).
@achow101 [this](https://github.com/Coldcard/firmware/pull/91) will also remove the need for `coldcard-libngu.path` in HWI (I will prepare PR when ready)

PR #97, opened 2022-05-06
Allow PSBTs with missing UTXOs

straylight-orbit:
It appears that the firmware is unable to sign PSBTs where one or more inputs lack a UTXO. This is a problem in multi-party transactions (e.g. coinjoin) where each signer is expected to populate their own UTXO data and ignore other inputs for signing purposes.

This change simply deletes the line that raises a fatal error in that scenario and uncomments an older code block that creates a warning instead.

### DIFF
--- a/shared/psbt.py
+++ b/shared/psbt.py
@@ -623,7 +623,6 @@ class psbtInputProxy(psbtProxy):
 
         return utxo
 
-
     def determine_my_signing_key(self, my_idx, utxo, my_xfp, psbt):
         # See what it takes to sign this particular input
         # - type of script
@@ -1249,7 +1248,7 @@ class psbtObject(psbtProxy):
         # Look an the UTXO's that we are spending. Do we have them? Do the
         # hashes match, and what values are we getting?
         # Important: parse incoming UTXO to build total input value
-        missing = 0
+        foreign = 0
         total_in = 0
 
         for i, txi in self.input_iter():
@@ -1258,9 +1257,13 @@ class psbtObject(psbtProxy):
                 self.presigned_inputs.add(i)
 
             if not inp.has_utxo():
-                # maybe they didn't provide the UTXO
-                missing += 1
-                continue
+                if inp.subpaths and not inp.fully_signed:
+                    # we cannot proceed if the input is ours and there is no UTXO
+                    raise FatalPSBTIssue('Missing own UTXO(s). Cannot determine value being signed')
+                else:
+                    # input clearly not ours
+                    foreign += 1
+                    continue
 
             # pull out just the CTXOut object (expensive)
             utxo = inp.get_utxo(txi.prevout.n)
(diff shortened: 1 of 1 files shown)


doc-hex:
I think this could be allowed, assuming we don't need to sign that input. However, it seems risky to allow otherwise, since PSBT tampering could result in loss of funds to huge miner fees... A warning is not enough, IMHO, in that case.

straylight-orbit:
My understanding is that loss of funds due to miner fees can happen only when output amounts are tampered with (i.e. when the sum of outputs is made to be less than the sum of inputs). In other words, providing bad amounts here will simply result in an invalid transaction being signed in the case of segwit (won't validate) and nothing special happening in the case of non-segwit inputs (legacy sighash doesn't include amounts). Either way, change is just another typical output so it's up to the PSBT provider to determine what the mining fee will be. Coldcard cannot be tricked into paying an unreasonably large fee since it doesn't craft outputs, it only signs them.

straylight-orbit:
Basically even if non-signing UTXOs are provided it still doesn't help since they aren't covered by our signature.

But to mitigate possible concerns, we could split the `missing` variable into two: `missing_ours` and `missing_other`. We'd still raise an error if `missing_ours > 0` and add a warning if `missing_other > 0`. Would that work?

straylight-orbit:
I think it's not about sighash types since amount+scriptpubkey (also known as `TxOut`) belonging to foreign (i.e. not our) inputs aren't covered even with `SIGHASH_ALL`. Only txids+vouts (also known as `Outpoint`) are.

In fact, you could populate those foreign `TxOut`s (or `UTXO`s as they're called here) with just about anything and the Coldcard will still correctly sign the transaction since it ignores non-signing `UTXO`s. I just don't think doing that is strictly correct, and clients shouldn't have to populate foreign inputs with dummy data if there's no intention/ability to sign them (lack of bip32 derivation info means no signature is produced). Since we are skipping signing foreign inputs altogether (as it should be), we may as well drop the requirement to supply their UTXO data.

Hence I think my original proposal doesn't introduce anything adverse. It strictly makes the signing logic more correct.

doc-hex:
> Since we are skipping signing foreign inputs altogether (as it should be), we may as well drop the requirement to supply their UTXO data.

Right. That's what I was missing.

scgbckbone:
hey @straylight-orbit, thanks for this PR. I still have some concerns - but cannot even describe them. Do you have some resources/bookmarks for me?, want to verify/understand this (was this discussed somewhere, is this implemented in some wallet, etc...), your reasoning seems sound, but I do not feel qualified enough here. Thanks again.

shared/psbt.py
```diff
@@ -1258,9 +1257,13 @@ def consider_inputs(self):
                 self.presigned_inputs.add(i)
 
             if not inp.has_utxo():
-                # maybe they didn't provide the UTXO
-                missing += 1
-                continue
+                if inp.subpaths and not inp.fully_signed:
```
scgbckbone, line 1260: not sure why are you checking for presence of `inp.subpaths` when you try to decide if input is ours? You can create valid PSBT which has foreign subpaths but lacks foreign UTXO

this better:
`if inp.num_our_keys and not inp.fully_signed:`

scgbckbone:
I tested with bitcoin core (v22) and it also does not care if there is UTXO for foreign inputs. It just happily signs even without foreign inputs UTXOs.

scgbckbone:
we will polish this a bit add some tests and give you heads up in release notes - thanks again - is it OK to mention you github nickname there?

scgbckbone:
https://github.com/Coldcard/firmware/pull/120

PR #98, opened 2022-05-07
USB Encryption v2

DON-MAC-256:
This introduces a new `ncry` version to close a potential attack vector:

A malicious program may re-initialize the connection encryption by sending the `ncry` command a second time during USB operation. This may prove particularly harmful in HSM mode.

Sending version `0x2` sets the new `bound` flag on the `USBHandler`, which changes the behavior in two ways:

- All future commands **must** be encrypted
- Returns an error if the `ncry` command is sent again for the duration of the power cycle

### DIFF
--- a/shared/usb.py
+++ b/shared/usb.py
@@ -110,6 +110,9 @@ class USBHandler:
 
         self.encrypted_req = False
 
+        # not bound to a specific crypto setup by default
+        self.bound = False
+
         # these will be objects later
         self.encrypt = None
         self.decrypt = None
@@ -167,6 +170,9 @@ class USBHandler:
 
                 if not(4 <= msg_len <= MAX_MSG_LEN):
                     raise FramingError('badsz')
+
+                if not is_encrypted and self.bound:
+                    raise FramingError('must encrypt')
 
                 if is_encrypted:
                     if self.decrypt is None:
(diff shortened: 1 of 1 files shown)


doc-hex:
Thanks for this well-thought-out addition!  Really puts a big reliability burden on the other end of the USB connection. As you say, probably only suitable for HSM mode. Any comms trouble and the Coldcard will need a reboot to recover USB operation. 

How would you like to be credited in the release notes? Twitter handle perhaps?

DON-MAC-256:
> Any comms trouble and the Coldcard will need a reboot to recover USB operation.

436aeb2ee3c094a5261d47d9c5bcbf2b260e1f4b adds a facility to shutdown the device -- exiting HSM -- in the event that an unknown command is received while the device is bound to the USB connection. This strengthens `ncry 0x2` specifically for the HSM case.

doc-hex:
Looks good. Last call before I merge...

DON-MAC-256:
Is this backported to Mk3 hardware?

doc-hex:
It was done for Mk4 only. I strongly recommend HMS users to upgrade to Mk4 because of the greater ram and larger txn signing capabilities.

DON-MAC-256:
> It was done for Mk4 only. I strongly recommend HMS users to upgrade to Mk4 because of the greater ram and larger txn signing capabilities.

That's understandable, but it would be helpful to have this backported for development and testing purposes.

PR #99, opened 2022-05-10
Fix repro build mk3

scgbckbone:
Issue: `make repro` fails trying to get incorrect version (5.0.3) from coinkite downloads for MK3
solution: hardocde MK3 version (4.1.5) into MK3-Makefile


doc-hex:
No, it's  much more than the version string. We are in an uncomfortable time of two major code branches.

PR #100, opened 2022-05-10
Image font

scgbckbone:
Issue: Courier font is not available on Ubuntu20. 
Proposed solution: use default font, that is bit less readable (but still pretty good imo [check attached picture]). Main advantage of default font is that it will be available on every machine, because it is base64 encode, in file in library (check site-packages/PIL/ImageFont.py)


scgbckbone:
font

scgbckbone:
will be part of different PR

PR #101, opened 2022-05-10
Unlink to remove

scgbckbone:
Issue: `uos.unlink` is not available (just linux?)
potential solution: 'uos.remove' 
Screenshot from 2022-05-10 11-57-08


scgbckbone:
not sure about availability of `uos.remove` on mac

PR #103, opened 2022-05-11
Safer HSM stat formatting

ghost:
In the HSM layout there is enough space for four digits for approvals and rejections. That could break the screen layout in the case of very long running HSM modes with approval/rejection numbers higher than 9999.

New formatting examples:
- 9,999 -> `9999`
- 10,000 -> `10K`
- 999,999 -> `999K`
- 542,345,451 -> `542M`
- 1,000,000,000 -> `1G+`

### DIFF
--- a/shared/hsm_ux.py
+++ b/shared/hsm_ux.py
@@ -234,7 +234,7 @@ class hsmUxInteraction:
 
         # 3 statistics; see draw_background for X positions
         y = 28+1
-        for x, val in [ (14, str(hsm_active.approvals)),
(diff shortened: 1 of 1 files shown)


doc-hex:
- from `hsm.py:27`
```
# too many refusals will cause reset
ABSOLUTE_MAX_REFUSALS = const(100)
```
- but yes, approvals could overflow display space
- more useful to see least-significant digits...
- showing `0001` for `10001` would kinda be better... could just add `[:4]` to `str(hsm_active.approvals))`

thoughts?

PR #104, opened 2022-05-18
Linux test compat

scgbckbone:
* add `run_sim_tests.py` as convenience test runner (against simulator)
* fix tests
* all test passing now (more comments in code)


testing/conftest.py
```diff
@@ -986,8 +987,7 @@ def doit(f_or_data, accept=True, finalize=False, accept_ms_import=False, complet
         if 'Choose PSBT file' in story:
             need_keypress('y')
             time.sleep(.1)
-            
-        pick_menu_item(psbtname+'.psbt')
+            pick_menu_item(psbtname+'.psbt')
```
scgbckbone, line 990: this was failing, you can only pick psbt file if 'Choose PSBT file' in story

testing/devtest/backups.py
```diff
@@ -20,7 +20,7 @@
     blanks = 0
     checklist = set('mnemonic chain xprv xpub raw_secret fw_date fw_version fw_timestamp serial '
                 'setting.terms_ok setting.idle_to setting.chain'.split(' '))
-    optional = set('setting.pms setting.axi setting.nick setting.lgto setting.usr hsm_policy setting.words long_secret multisig setting.multisig setting.fee_limit setting.tp setting.check duress_xprv duress_xpub duress_1001_words duress_1002_words duress_1003_words'.split(' '))
+    optional = set('setting.nfc setting.pms setting.axi setting.nick setting.lgto setting.usr hsm_policy setting.words long_secret multisig setting.multisig setting.fee_limit setting.tp setting.check duress_xprv duress_xpub duress_1001_words duress_1002_words duress_1003_words'.split(' '))
```
scgbckbone, line 23: added to `setting.nfc` optional

testing/requirements.txt
```diff
@@ -13,7 +13,8 @@ zbar-py==1.0.4
 
 # NFC and NDEF handling
 nfcpy==1.0.3
-ndef==0.2
+# commentt out below dependency 'ndef==0.2' if on debian based OS
+#ndef==0.2
```
scgbckbone, line 17: not sure if this is needed in MacOS. But when I install nfc.py it also pulls as a dependency ndeflib which is what we want. ndef==0.2 was obsolete and does not work

testing/test_hsm.py
```diff
@@ -1143,9 +1127,11 @@ def test_boot_to_hsm_unlock(start_hsm, hsm_status, enter_local_code):
     enter_local_code('123123')
     time.sleep(.5)
     assert hsm_status().active == False
-    assert hsm_status().policy_available == False
+    assert hsm_status().policy_available == True  # we haven't removed anythong why shoudl it be not available?
```
scgbckbone, line 1130: Not sure why this was False before, but was not working --> maybe I'm missing something here ?

scgbckbone:
`test_multisig.py::test_ms_sign_simple` takes 40min and imo should be moved to @veryslow

Currently full run (`python run_sim_tests.py -m all --onetime --veryslow`) --> cca 235 minutes
Run without @veryslow and @onetime (`python run_sim_tests.py -m all `) --> 90 minutes
if `test_multisig.py::test_ms_sign_simple` moved to @veryslow above would be only around 50 minutes (lunch break time?)

doc-hex:
> `test_multisig.py::test_ms_sign_simple` takes 40min and imo should be moved to @veryslow

Sure!

shared/utils.py
```diff
@@ -102,7 +102,7 @@ def __enter__(self):
         return self
 
     def __exit__(self, *a, **k):
-        self.fd.seek(0, 3)          # go to end
+        self.fd.seek(0, 2)          # go to end
```
doc-hex, line 105: this would have been a bug; is there a test case that triggers/demos the bug before/after fix?

testing/pytest.ini
```diff
@@ -1,12 +1,14 @@
...
-#addopts = -vv
+# you need to comment above and uncomment below to use run_sim_tests.py
+#addopts = -vv --disable-warnings
 markers = 
     bitcoind: indicates local bitcoind (testnet) will be needed
     onetime: test cant be combined with any others, likely needs board reset
     veryslow: test takes more than 30 minutes realtime
     qrcode: test uses or tests QR related features
     unfinalized: test cases produces an unfinalized PSBT
+    manual: test cannot be combined with nay others, check for "fully done" in repl (then it will hang - kill it)
```
doc-hex, line 11: nay -> any

testing/requirements.txt
doc-hex: `ndeflib` in install requirements of nfcpy so okay to remove; kill comment too imho

testing/run_sim_tests.py
```diff
@@ -0,0 +1,258 @@
...
+import os
+import time
+import glob
+import json
+import pytest
+import atexit
+import signal
+import argparse
+import subprocess
+import contextlib
```
doc-hex, line 37: - merge these imports, add Copyright to top of file
- would prefer `click` to `argparse`, but not worth changing?

testing/run_sim_tests.py
```diff
@@ -0,0 +1,258 @@
...
+        for fn in args.module:
+            if not os.path.exists(fn):
+                raise RuntimeError(f"{fn} does not exist")
+        test_modules = sorted(args.module)
+    result = []
+    for test_module in test_modules:
+        test_args = DEFAULT_SIMULATOR_ARGS
+        print("Started", test_module)
+        if test_module in ["test_rng.py", "test_pincodes.py"]:
+            # test_pincodes.py can only be run against real device
```
doc-hex, line 217: this should be achieved via `pytest.skip` or `xfail` inside the test code

doc-hex: 2nd thought: good this way, like other special cases per-file.

testing/test_hsm.py
doc-hex: I think an earlier version of the firmware always deleted the HSM policy file when unlocked in this manner, but then that got removed (since hard to reload policy) and so this test became wrong.

shared/utils.py
doc-hex: - looks like `seek(0,3)` was a noop based on my reading of `external/micropython/extmod/vfs_fat_file.c:112`
- so change looks like good fix of a bug that never had impact

doc-hex:
Looking good. Last call before I merge.

PR #114, opened 2022-06-06
Update and rename menu-tree.txt to menu-tree-spanish.txt

JaimeWGarcia:
If you need a translation of your Mk4 menu to Spanish. I made a suggestion.

---


doc-hex:
- I don't think we want to delete the existing English file.
- English file is auto-generated and so that's why it's here, but a translated version belongs on the docs website not Github
- I think side-by-side would be more useful, since the UX on the Coldcard remains in English.

JaimeWGarcia:
Thanks. My first time using GitHub. The idea was that there was a language option and that there was a parallel menu tree translated into Spanish and not to replace the English menu. Anyway, if you're ever interested in the translation of the menu tree, it is already done and you may use it if you wish. All the best.

PR #121, opened 2022-06-21
allow unknown scripts in HSM; add support for OP_RETURN

scgbckbone:
* Allow unknown scripts in HSM mode (same approach as `render_output`)
* OP_RETURN is no longer unknown script and during signing, if all elements in op_return data are in range 32 <= elem<127 (printable) will be displayed in ascii after hex representation
*  testing with bitcoind


shared/hsm.py
```diff
@@ -759,7 +759,10 @@ async def approve_transaction(self, psbt, psbt_sha, story):
                 for idx, tx_out in psbt.output_iter():
                     if not psbt.outputs[idx].is_change:
                         total_out += tx_out.nValue
-                        dests.append(chain.render_address(tx_out.scriptPubKey))
+                        try:
+                            dests.append(chain.render_address(tx_out.scriptPubKey))
+                        except ValueError:
+                            dests.append(str(b2a_hex(tx_out.scriptPubKey), 'ascii'))
```
doc-hex, line 765: This must render OP_RETURN and raise warning if we can't understand the destination. [Maybe a policy flag (default off) that permits sending to unknown scripts—but no one is asking for that.]
We need to refactor `auth.py: render_output()` into either `psbt.py` or `chains.py`

doc-hex: Hmmm.. altho this is just for the policy engine. IDK

PR #122, opened 2022-06-21
multisig descriptor

scgbckbone:
* Add ability to import/export multisig wallet as descriptor. 
* with descriptor wallets, ~~support both multi and sortedmult~~ only sortedmulti descriptors are supported
* provided 2of2 tutorial with bitcoind and descriptors and 2of2 (two coldcards) and bitcoin-qt as coordinator
* tests

Limitations:
* only support xpub ranged descriptors (enough for our use case)
* only external descriptor (this needs discussion, should we rather import both, etc...)

Known issues:
* once you imported wallet from descriptor, you probably cannot export the same wallet for electrum (if you do not have 0/0* sub derivations) --> this will need some warning
* NFC untested


scgbckbone:
reworking, simplifying, etc. ---> will provide different PR

PR #125, opened 2022-07-09
Patch: Increase scan rate of ssd1306 display.

russeree:
### Why Increase Display Scan Rate?

1. The primary reason to fix this is because of the fact that QR codes are borderline impossible to scan in any bright light. This is because as the shutter speed of mobile camera increases to compensate for bright environments. The ssd1306 oled display can't complete a full scan during the time the time the camera shutter is open. Increasing the scan rate remedies this to a degree. This means the QR code will be unreadable with default exposure settings in bright light with most phones. Issues begin when a cameras shutter speed is faster than 1/60 of a second. To test this use 'pro' mode on any modem phone and adjust the shutter speed below and above 1/60 of a second to see the effect side by side. 

2. Reduced eyestrain and flicker, since the display is completing scans nearly 2x as fast the display appears to flicker less even during normal user. 

### What do these QR scanning failures look like?
image

### Mechanism of Fix
Writing a 0xF to d7-d4 of register D5. Check the table below. This increases the scan rate of the display. The patch is non invasive and doesn't induce an addition attack surface since it's just a parameter change. This wont remedy partial images when photographed or scanned by cellphones when the shutter speed is faster than 1/120 sec.

image


scgbckbone:
why did you removed `file_time.c` in those two locations ?

scgbckbone:
_QR codes are borderline impossible to scan in any bright light._ --> this is unfortunately true...

Just tested, your fix is solving the issue. QR can now be scanned in spaces with bright light.

Below is a video (left current, right this branch)

https://user-images.githubusercontent.com/25349625/178103279-b0f4a46e-62b8-448c-84d9-3e2526d07d82.mp4

ACK

doc-hex:
Nice fix!

I want to measure the current before/after this change and see what the impact is. If it's bad, we can still make the change, but we'll do it only while showing a QR on screen.

russeree:
> why did you removed `file_time.c` in those two locations ?

I was trying to keep the commit clean, and the files were "autogenerated", I will fix these issues and squash this evening. 


> Nice fix!
> 
> I want to measure the current before/after this change and see what the impact is. If it's bad, we can still make the change, but we'll do it only while showing a QR on screen.

Thanks there is also a second way that could be used to improve QR readability that would require more work potentially. I will test this weekend. That would be to use vertical addressing mode on the display. Currently there is a guarantee that if the phones camera shutter rate is faster than the scan freq. of the display that there will be some level of 'tearing'. Sorry I don't have a better work on hand at the moment. 

switching to horizontal addressing would remedy this further because the scan would happen left to right and the tear when seen by the phones camera would only exist at some point on top of the text next to the qr code. Not the qr code it self. The QR code takes up 100% of the vertical space but only 50% of the horizontal space. Any tear vertically obstructs the qr code, a horizontal tear is not guaranteed. 

image

Also lastly do qr codes generated use 'M' level ECC?

PR #132, opened 2022-07-22
tutorial migrated to coldcard.com

scgbckbone:
* remove bitcoin-qt tutorial (link instead pointing to coldcard.com) from firmware repo
* kept bitcoind tutorial as it is only .md (no pics)


doc-hex:
I dont like links in section titles. Just do 

```
Full tutorial [now on coldcard.com/docs](deeplink)

PR #134, opened 2022-08-02
add proof test - order of xpubs in multisig does not matter

scgbckbone:


doc-hex:
test would be more conclusive if it did all possible orderings

scgbckbone:
all options tested now

doc-hex:
now that's a test!

PR #139, opened 2022-08-21
New HSM policy rules

straylight-orbit:
The PR contains a number of additions to the HSM policy format.

There is a new rule called `min_pct_self_transfer`. This operates on own inputs and outputs, where "own" means "contains a derivation path" and can therefore be guaranteed to derive from the master key on the device. The rule accepts a percentage (0..100 float) and calculates the ratio of funds going to own outputs and funds coming from own inputs. For instance, `"min_pct_self_transfer": 99.0` would dictate that the value of outputs going back to the wallet has to meet at least 99% of the value of inputs coming from the wallet  (in other words, non-change outputs are confined to 1% or lower). Foreign inputs and outputs are ignored for the purposes of this rule.

Another addition is the `patterns` list which takes any number of predefined patterns. These patterns are rather simple heuristics that are applied against a PSBT. The following patterns have been defined thus far:

```json
{
    "EQ_NUM_INS_OUTS": "the number of inputs and outputs must be equal",
    "EQ_NUM_OWN_INS_OUTS": "the number of OWN inputs and outputs must be equal",
    "EQ_OUT_AMOUNTS": "all outputs must have equal amounts"
}
```
The definition of "own" in the second pattern is the same as the one described earlier.

This rule might be used like: ```"patterns": ["EQ_NUM_INS_OUTS", "EQ_OUT_AMOUNTS"]```.

The addition of these building blocks allows the creation of policies that suit more complex transactions (such as Chaumian coinjoin) while blocking everything else and ensuring that malicious software cannot drain a Coldcard using fraudulently crafted transactions. The pattern list is extensible and can accommodate further applications.

### DIFF
--- a/shared/hsm.py
+++ b/shared/hsm.py
@@ -114,7 +114,6 @@ def pop_float(j, fld_name, mn=0, mx=1000000):
     if v is None: return v
     assert float(v) == v, "%s: must be float" % fld_name
     v = float(v)
-    assert mn <= mx, '%s: cannot be specified' % fld_name
     assert mn <= v <= mx, "%s: must be in range: [%d..%d]" % (fld_name, mn, mx)
     return v
 
@@ -306,8 +305,8 @@ class ApprovalRule:
         if self.min_pct_self_transfer:
             own_in_value = sum([i.amount for i in psbt.inputs if i.num_our_keys > 0])
             own_out_value = sum([o.amount for o in psbt.outputs if o.num_our_keys > 0])
-            ratio = float(own_out_value) / own_in_value
-            assert ratio >= self.min_pct_self_transfer, 'does not meet self transfer threshold'
+            percentage = (float(own_out_value) / own_in_value) * 100.0
+            assert percentage >= self.min_pct_self_transfer, 'does not meet self transfer threshold, expected: %.2f, actual: %.2f' % (self.min_pct_self_transfer, percentage)
 
         # check various patterns
 
--- a/shared/hsm.py
+++ b/shared/hsm.py
@@ -31,6 +31,13 @@ ABSOLUTE_MAX_REFUSALS = const(100)
 # mode, if you enable the boot_to_hsm feature
 BOOT_LOCKOUT_TIME = const(60)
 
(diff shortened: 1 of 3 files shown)


doc-hex:
Nice stuff! Here's my feedback:

- revert the new`amount` field on `psbt.py`
- instead, add a new function that calculates "own value" for ins and outs (return a tuple, and cache it as well)
- call that in `matches_transaction` as needed, remove list comprehensions
- [iterating over psbt ins and outs is super expensive, btw]

shared/hsm.py
```diff
@@ -266,6 +302,28 @@ def matches_transaction(self, psbt, users, total_out, dests, local_oked):
             # check this txn would not exceed the velocity limit
             assert self.spent_so_far + total_out <= self.per_period, 'would exceed period spending'
 
+        # check the self-transfer percentage
+        if self.min_pct_self_transfer:
+            own_in_value = sum([i.amount for i in psbt.inputs if i.num_our_keys > 0])
+            own_out_value = sum([o.amount for o in psbt.outputs if o.num_our_keys > 0])
+            ratio = float(own_out_value) / own_in_value
+            assert ratio >= self.min_pct_self_transfer, 'does not meet self transfer threshold'
```
doc-hex, line 310: add actual ratio to this error msg

doc-hex: and expected value, both as percents

doc-hex:
- forget wh

> Nice stuff! Here's my feedback:
> 
> * revert the new`amount` field on `psbt.py`
> * instead, add a new function that calculates "own value" for ins and outs (return a tuple, and cache it as well)
> * call that in `matches_transaction` as needed, remove list comprehensions
> * [iterating over psbt ins and outs is super expensive, btw]

... hmmm, altho what you had is actually less iterating over the inputs ... 

Forget what I said above. It's good as-is.

doc-hex:
- look good, just needs test cases
- look at `testing/test_hsm.py` and case `test_velocity()` for ideas

doc-hex:
Looks great! Thanks for these new features!

PR #142, opened 2022-08-29
PSBT and HSM refactor

straylight-orbit:
This PR is mainly a refactor along with some improvements. The changes are as follows:

1. OP_RETURN outputs that spend (burn) 0 sats are no longer subjected to whitelist testing. This is because they're harmless and don't consume value.

2. OP_RETURN tests used to require a running bitcoind instance. This is no longer the case because `fake_txn` has been updated to support arbitrary OP_RETURN data.

3. PSBT proxy objects were caching the `amount` value in `validate`. This wasn't the best place because that method is concerned with validating own outputs. This has been moved to `consider_outputs`. Scriptpubkey is now also cached there.

4. As a consequence of caching scriptpubkey in proxy objects, the expensive `output_iter` method which operates directly on the filesystem no longer needs to be called when doing whitelist checks. This should result in improved performance. There are probably other places that could benefit from a similar refactor.

5. Whitelist checks have been moved to a more appropriate place (`rule.matches_transaction`).

All tests have been run against the simulator and no regressions have been introduced.

### DIFF
--- a/testing/txn.py
+++ b/testing/txn.py
@@ -179,7 +179,7 @@ def render_address(script, testnet=True):
     # string... aka: the "payment address"
     from pycoin.encoding import b2a_hashed_base58
     from bech32 import encode as bech32_encode
-    from binascii import hexlify
+    from binascii import hexlify as b2a_hex
 
     ll = len(script)
 
@@ -212,7 +212,7 @@ def render_address(script, testnet=True):
 
     # OP_RETURN
     if script[0:1] == b'\x6a':
-        return hexlify(script)
+        return b2a_hex(script)
 
     raise ValueError('Unknown payment script', repr(script))
 
--- a/testing/txn.py
+++ b/testing/txn.py
@@ -63,7 +63,7 @@ def fake_txn(dev):
 
     def doit(num_ins, num_outs, master_xpub=None, subpath="0/%d", fee=10000,
                 invals=None, outvals=None, segwit_in=False, outstyles=['p2pkh'], psbt_hacker=None,
-                change_outputs=[], capture_scripts=None, add_xpub=None):
+                change_outputs=[], capture_scripts=None, add_xpub=None, op_return = None):
         psbt = BasicPSBT()
         txn = Tx(2,[],[])
         master_xpub = master_xpub or dev.master_xpub or simulator_fixed_xprv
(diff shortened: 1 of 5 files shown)


scgbckbone:
1. not sure about this - once user specify a whitelist - does he really wants to sign transaction with random number of OP_RETURN scripts ? "harmless" is understatement as you still have to pay fees for them
2. lgtm (yet I would still test with bitcoind so the fake_txn isn't the only OP_RETURN provider against which we test)
3.,4.,5. ok

scgbckbone (commented):
1. not sure about this - once user specify a whitelist - does he really wants to sign transaction with random number of OP_RETURN scripts ? "harmless" is understatement as you still have to pay fees for them
2. lgtm (yet I would still test with bitcoind so the fake_txn isn't the only OP_RETURN provider against which we test)
3.,4.,5. ok

scgbckbone:
in our test suite, for some tests fresh simulator is needed (one of them is `test_sign.py::test_bip143_attack_data_capture[True-100]`)

As we expect some tests to fail during the run - we have implemented our own test driver that replays failed test after each run against fresh simulator ---> check module doc string in `testing/run_sim_tests.py` (firmware).

scgbckbone:
Second thoughts - you're right with harmlessness (based on my knowledge). But IMO once you have a whitelist, it should only allow what is whitelisted - by definition.

If you want/need to allow random OP_RETURN scripts ?maybe? some new  policy option (default=do not allow random OP_RETURN) and not part of whitelist...

scgbckbone (changes requested):
above

scgbckbone (approved):
Tested ACK

scgbckbone:
You also need an ACK from @doc-hex to get this in. Just build on top of this - rebasing is easy...

testing/txn.py
```diff
@@ -164,6 +179,7 @@ def render_address(script, testnet=True):
     # string... aka: the "payment address"
     from pycoin.encoding import b2a_hashed_base58
     from bech32 import encode as bech32_encode
+    from binascii import hexlify
```
doc-hex, line 182: please use `b2a_hex` rather than poorly named `hexlify`

doc-hex:
Looking good, safe to build on... just need to satisfy my pet-peeve about `b2a_hex`

PR #143, opened 2022-09-03
HSM Dynamic Whitelisting

straylight-orbit:
This PR introduces a new HSM mode that enables dynamic whitelisting.

### Why is this needed?

While the Coldcard can tell which outputs belong to itself, blocking malicious outputs in HSM mode is difficult. This is normally solved with whitelisting. But whitelisting works only with static addresses and is not practical in dynamic applications where destination addresses are unknown ahead of time. Sometimes we want "dynamic" whitelisting where a PSBT can prove to the Coldcard that foreign outputs are legitimate.

### How is this achieved?

Outputs can be attested by trusted keys, where a trusted key creates a signature (attestation) over a serialized txout and thereby says "I am OK with this scriptpubkey and its amount". Each foreign output's attestation is attached to a custom per-output proprietary PSBT entry (proprietary keys format is defined in [BIP174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#proprietary-use-type)).

Public keys (addresses) used for verification are loaded into the HSM policy using the existing `whitelist` list.

### Usage
HSM policy example that uses attestation-based dynamic whitelisting:

```json
{
    "rules": [
        {
            "whitelist": [
                "17eet2ZwDT4PHvsYeTiWggMQUYfRfA3XYH",
                "1BQdJXHXao72kMWmDBppvEkR56EXxWSrnN"
            ],
            "whitelist_opts": {
                "mode": "ATTEST",
                "allow_zeroval_outs": true
            }
        }
    ]
}
```

`whitelist_opts` is a new structure that contains two fields:
- `mode: str`: `BASIC` or `ATTEST`. `BASIC` is the old behavior where whitelisting is based on address comparison. `ATTEST` is the new attestation mode.
- `allow_zeroval_outs: bool`: whether to always allow outputs with no value (OP_RETURN) without requiring attestation. This setting applies to either mode.

An output that has been attested will set the following key-value pair in its own PSBT entry.

- key:
-- keytype: ` 0xFC` -- stands for `proptietary`, defined by BIP174
-- identifier: `COINKITE` (global namespace used by Coldcard)
-- subtype: `0` (we define this as "attestation signature")
-- keydata: empty
- value:
-- 65 byte recoverable signature

### Verification process

A Coldcard with a policy designed to perform attestation-based whitelisting takes the following steps:

1. extracts attached attestation (signature) for each PSBT output entry
2. if output is foreign, its signature is verified and a public key is recovered from it
3. extracted public key is hashed, converted to an address and checked against addresses in the whitelist
4. if present in the whitelist pass, otherwise fail

### Testing

Existing tests are passing and no regressions have been introduced. New tests have been written that test the attestation mode by covering txouts with both valid and invalid signatures and assert the correct outcome. PSBT serialization/deserialization when attestation values are present has also been covered. Legacy OP_RETURN whitelist tests have been updated to take `allow_zeroval_outs` into account.

@doc-hex @scgbckbone

### DIFF
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -5,7 +5,7 @@
 #
 import stash, ure, ux, chains, sys, gc, uio, version, ngu
 from public_constants import MSG_SIGNING_MAX_LENGTH, SUPPORTED_ADDR_FORMATS
-from public_constants import AFC_SCRIPT, AF_CLASSIC, AFC_BECH32, AF_P2WPKH
+from public_constants import AFC_SCRIPT, AF_CLASSIC, AFC_BECH32, AF_P2WPKH, AF_P2WPKH_P2SH
 from public_constants import STXN_FLAGS_MASK, STXN_FINALIZE, STXN_VISUALIZE, STXN_SIGNED
 from sffile import SFFile
 from ux import ux_aborted, ux_show_story, abort_and_goto, ux_dramatic_pause, ux_clear_keys
@@ -132,6 +132,28 @@ RFC_SIGNATURE_TEMPLATE = '''\
 {sig}
 -----END {blockchain} SIGNED MESSAGE-----
 '''
+
+def verify_recover_pubkey(sig, digest):
+    # verifies a message digest against a signature and recovers
+    # the address type and public key that did the signing
+    if len(sig) != 65:
+        raise ValueError('signature length != 65')
+
+    v = sig[0]
+    if 27 <= v <= 34:
+        af = AF_CLASSIC
+    elif 35 <= v <= 38:
+        af = AF_P2WPKH_P2SH
+    elif 39 <= v <= 42:
+        af = AF_P2WPKH
+    else:
+        raise ValueError('unsupported recovery id v=%s' % v)
+
+    try:
+        sig = ngu.secp256k1.signature(sig)
+        return af, sig.verify_recover(digest).to_bytes()
+    except:
+        raise ValueError('invalid signature')
 
 def sign_message_digest(digest, subpath, prompt):
     # do the signature itself!
--- a/shared/psbt.py
+++ b/shared/psbt.py
@@ -26,6 +26,15 @@ from public_constants import (
     PSBT_OUT_BIP32_DERIVATION, MAX_PATH_DEPTH
 )
 
+# PSBT proprietary keytype
+PSBT_PROPRIETARY = const(0xFC)
+
+# PSBT proprietary identifier for Coinkite applications
+PSBT_PROP_CK_ID = b"COINKITE"
+
+# PSBT proprietary subtype for attestation entries
+PSBT_ATTESTATION_SUBTYPE = const(0)
+
 # Max miner's fee, as percentage of output value, that we will allow to be signed.
 # Amounts over 5% are warned regardless.
 DEFAULT_MAX_FEE_PERCENTAGE = const(10)
@@ -47,18 +56,6 @@ class HashNDump:
     def digest(self):
         print(' END')
         return self.rv.digest()
-
-def read_varint(v):
-    # read "compact sized" int from a few bytes.
-    assert not isinstance(v, tuple), v
-    nit = v[0]
-    if nit == 253:
-        return unpack_from("<H", v, 1)[0]
-    elif nit == 254:
-        return unpack_from("<I", v, 1)[0]
-    elif nit == 255:
-        return unpack_from("<Q", v, 1)[0]
-    return nit
 
 def seq_to_str(seq):
     # take a set or list of numbers and show a tidy list in order.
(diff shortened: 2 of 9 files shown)


shared/psbt.py
```diff
@@ -26,6 +26,15 @@
     PSBT_OUT_BIP32_DERIVATION, MAX_PATH_DEPTH
 )
 
+# PSBT proprietary keytype
+PSBT_OUT_PROPRIETARY = const(0xFC)
```
scgbckbone, line 30: why is this OUT? isn't proprietary in global namespace?

shared/auth.py
```diff
@@ -133,6 +133,17 @@ async def failure(self, msg, exc=None, title='Failure'):
 -----END {blockchain} SIGNED MESSAGE-----
 '''
 
+def verify_recover_message_digest(sig, digest):
```
scgbckbone, line 136: `verify_recover_pubkey` ?? --> maybe this method could also return address format (AF_<what>) based on parsed header byte?

scgbckbone (changes requested):
conceptACK (I like it very much as it is making whitelisting sooo much more flexible)

* no guessing of address type --> use header byte to decide addr format

scgbckbone:
| y-parity | x-order | recovery id | v | addr type | addr prefix |
|:--------:|:-----------:|:-----------:|:--:|:----------------------:|:------------------:|
| 0 | less than n | 0 | 27 | uncompressed p2pkh | 1, m, n |
| 1 | less than n | 1 | 28 | uncompressed p2pkh | 1, m, n |
| 0 | more than n | 2 | 29 | uncompressed p2pkh | 1, m, n |
| 1 | more than n | 3 | 30 | uncompressed p2pkh | 1, m, n |
| 0 | less than n | 0 | 31 | compressed p2pkh | 1, m, n |
| 1 | less than n | 1 | 32 | compressed p2pkh | 1, m, n |
| 0 | more than n | 2 | 33 | compressed p2pkh | 1, m, n |
| 1 | more than n | 3 | 34 | compressed p2pkh | 1, m, n |
| 0 | less than n | 0 | 35 | compressed p2sh-p2wpkh | 3, 2 |
| 1 | less than n | 1 | 36 | compressed p2sh-p2wpkh | 3, 2 |
| 0 | more than n | 2 | 37 | compressed p2sh-p2wpkh | 3, 2 |
| 1 | more than n | 3 | 38 | compressed p2sh-p2wpkh | 3, 2 |
| 0 | less than n | 0 | 39 | compressed p2wpkh | bc1q, tb1q, bcrt1q |
| 1 | less than n | 1 | 40 | compressed p2wpkh | bc1q, tb1q, bcrt1q |
| 0 | more than n | 2 | 41 | compressed p2wpkh | bc1q, tb1q, bcrt1q |
| 1 | more than n | 3 | 42 | compressed p2wpkh | bc1q, tb1q, bcrt1q |
| does NOT | support | recovery | 46 | compressed p2tr | bc1p |


`v` in table is a header byte - this is the table which I created as a response to https://github.com/ZenulAbidin/bips/blob/master/bip-notatether-signedmessage.mediawiki which has some typo/bugs there - this is how I think author intended to specify it so we do not fork BIP137

testing/helpers.py
```diff
@@ -198,4 +198,41 @@ def slip132undo(orig):
...
+        adjust = 0
+    elif addr_fmt == AF_P2WPKH_P2SH:
+        adjust = 4
+    elif addr_fmt == AF_P2WPKH:
+        adjust = 8
+    else:
+        raise ValueError('unsupported address format for signing')
+
+    if adjust:
+        # Python doesn't support byte assignment so make a list first...
```
scgbckbone, line 228: I'd prefer it this way (sorry for cosmetic comments):
`bytes([sig[0] + adjust]) + sig[1:]`

shared/psbt.py
```diff
@@ -48,18 +57,6 @@ def digest(self):
         print(' END')
         return self.rv.digest()
 
-def read_varint(v):
```
scgbckbone, line 51: why deleted? unused ?

scgbckbone (approved):
this attestation mode is lit! I was thinking about something similar, but I wanted to go down the way of xpubs and derivation - however this is much better - thank you! now let's get @doc-hex on board

doc-hex:
I also feel this is "lit". Will review and merge.

scgbckbone:
please squash the commits

scgbckbone:
Could you also add short but descriptive sentence to `releases/ChangeLog-mk4.md`?

scgbckbone (approved):

releases/ChangeLog-mk4.md
```diff
@@ -9,6 +9,8 @@
 - Enhancement: Block HSM (user management commands and HSM related commands) by default.
   To enable and use HSM go to Advanced/Tools -> Enable HSM -> Enable.
 - Enhancement: Allow import of new descriptor type which specify both internal/external in single string
+- Enhancement: Dynamic whitelisting in HSM mode is now possible. Foreign outputs can
```
scgbckbone, line 12: Enhancement: HSM dynamic whitelisting. Foreign outputs can be attested by being signed with a private key corresponding to the address specified in `whitelist` section. Attestation signature MUST be provided in PSBT.

?

scgbckbone (changes requested):

scgbckbone (approved):
Thanks

PR #144, opened 2022-09-05
Block User management commands and HSM related commands if not explicitly enabled

scgbckbone:


doc-hex:
afterthought: maybe we should enable these commands automatically, when an HSM config is imported?

scgbckbone:
updated

doc-hex:
Question: Will this work w/ "boot to HSM" feature and/or system where HSM was in use in previous firmware version?
I think we should force the new config value to True, if evidence HSM was on previous during upgrade process or first start after upgrade.

scgbckbone:
@straylight-orbit I would consider 1. and 2. to be just one step with confirmation. With my latest addition if your policy file is already there - you do not need to do any extra step after you upgrade to version with this PR code

PR #151, opened 2022-09-16
Display HSM policy hash on first activation

straylight-orbit:
Large and complex HSM policies can be difficult for users to visually parse and fully comprehend.

This UX improvement makes it possible to compute the canonical hash of a policy and display that hash on the secondary confirmation (policy installation) screen. One scenario where this is interesting is when an organization wants its remote members to install a policy but can't rely on their correctly visually parsing the policy and understanding it. In that case, the organization can send allowed policy hashes through a secondary communication medium and instruct its members to only install a policy with a matching hash. Policy hashes are canonical and don't depend on the ordering of keys, white spaces etc.

The same effect couldn't be achieved with just fingerprinting since it's trivial to create a fingerprint collision through grinding.

The confirmation screen now looks like this:
hash_1
hash_2

Existing tests are all passing and no regressions have been introduced.


testing/test_hsm.py
```diff
@@ -356,6 +356,7 @@ def doit(policy):
 
             title, body2 = cap_story()
             assert 'Last chance' in body2
+            assert 'Policy hash:' in body2
```
doc-hex, line 359: plz pull out hash (use a regex) and verify the math

doc-hex:
Interesting and useful feature, thanks!

testing/test_hsm.py
```diff
@@ -48,6 +48,52 @@
...
 ]
 
+def compute_policy_hash(policy):
+    # Computes a policy hash for comparison purposes. This has to do some conversion and cleanup
+    # in order to be able to see a policy the same way the firmware sees it.
+    hasher = sha256()
+    top_keys = ['must_log', 'never_log', 'priv_over_ux', 'warnings_ok', 'msg_paths',
+                    'share_xpubs', 'share_addrs', 'notes', 'period', 'allow_sl', 'set_sl', 'boot_to_hsm']
+    for key in top_keys:
+        if value := policy.get(key):
```
scgbckbone, line 58: `:=` this forks off all < py3.8 from running tests (I personally do not care as I already use py3.8 but have no idea about "others")

scgbckbone (approved):
tested ACK

testing/test_hsm.py
doc-hex: yeah that's ok, keep it

doc-hex:
- I think I'd prefer the hash was `sha256(ujson.dumps(self.save()))`
- handles future additions w/o changes
- `HSMPolicy.save()` return value is not deterministic tho, and we have no `OrderedDict()` on this platform
- same for `ApprovalRule.to_json()` and `WhitelistOpts.to_json()`
- could make those output json-serialized strings instead of objects for `ujson.dump` .. so rendering json manually
- or just make a basic `OrderedDict` [search first since someone else will have done already for micropython]

shared/hsm.py
```diff
@@ -177,7 +179,12 @@ def __init__(self, from_dict):
 
     def to_json(self):
         flds = [ 'mode', 'allow_zeroval_outs' ]
-        return dict((f, getattr(self, f, None)) for f in flds)
+        rv = OrderedDict()
+        for f in flds:
+            val = getattr(self, f, None)
+            if val:
```
scgbckbone, line 185: this would ignore both `None` and `False` --> does not seem right ignoring values

scgbckbone (changes requested):
do not ignore values

PR #154, opened 2022-10-04
Ephemeral seeds, final touches

doc-hex:
My changes to @scgbckbone branch


scgbckbone:
great! squashed and remove merge commit to this https://github.com/Coldcard/firmware/pull/152

PR #157, opened 2022-10-13
Manual steps to verify reproducible builds

i5hi:
We have been attempting to manually verify reproducibility of Coldcard firmware builds and lay out a process by which anyone can get an exact match on the shasum without any diff in the builds resulting from `make repro` in comparison to the release candidates found at https://coldcard.com/downloads/.

Following up on the discussions from twitter, as I understand, the diffs exist because of the signatures in the binary and (I think) some bootloader stuff. 

`stm32/shared.mk` contains code under `check_repro` which seems to do this.

Opening this PR so we can understand what is happening better here than twitter.

### DIFF
--- a/docs/notes-on-repro.md
+++ b/docs/notes-on-repro.md
@@ -1,0 +1,180 @@
+# Notes on Reproducible Builds
+
+The following document aims to breakdown how reproducibility is verified in `make repro` build process.
+
+## stm32/shared.mk
+
+The entrypoint makefile for repro builds.
+
+### repro
+
+The `repro` command in `shared.mk` is the first step in the repro build process, which triggers a docker build and run process.
+
+```makefile
+repro: 
+	docker build -t coldcard-build - < dockerfile.build
+	(cd ..; docker run $(DOCK_RUN_ARGS) sh src/stm32/repro-build.sh $(VERSION_STRING) $(MK_NUM))
+```
+
+Below are interesting sections from the docker logs that give an idea as to what is going on in build process:
+
+```stdout
++ mkdir /tmp/checkout
++ mount -t tmpfs tmpfs /tmp/checkout
+
+...
+
++ git clone /work/src/.git firmware
+
+...
+
++ cd firmware/external
++ git submodule update --init
+
+...
+
+Successfully installed signit-1.0
+
+...
+
++ cd ../stm32
++ cd ../releases
++ '[' -f '*-v5.0.7-mk4-coldcard.dfu' ]
++ dd 'bs=66' 'skip=1'
++ grep -F v5.0.7-mk4-coldcard.dfu signatures.txt
+0+1 records in
+0+1 records out
++ PUBLISHED_BIN=2022-10-05T1724-v5.0.7-mk4-coldcard.dfu
++ '[' -z 2022-10-05T1724-v5.0.7-mk4-coldcard.dfu ]
++ wget -S https://coldcard.com/downloads/2022-10-05T1724-v5.0.7-mk4-coldcard.dfu
+
+...
+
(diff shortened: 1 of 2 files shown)


i5hi:
Unable to tag @moonsettler who has been also been attempting this.

doc-hex:
- please make a whole new file in `./docs/` since the README is getting too big (then link to it)
- I suggest working from the above makefile snippet, and just explaining what each line does in detail

docs/notes-on-repro.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+To summarize `check-repro`:
+
+- At the final `check-repro` step, we have a locally built `firmware-signed.bin` and we want to check that it matches 
+the binary release provided by Coldcard.
+- An identical checksum match will not be possible as is, since there is signature data embedded into into the binary, which must be removed.
+- The specific release of the version that is being built is fetched, and placed it under /tmp/checkout/firmware/releases/*.dfu
+- `split` (cli/signit.py: Line 153-175) is run against the release `*.dfu` resulting in a `check-fw.bin` and `check-bootrom.bin`
+- `check` (cli/signit.py: Line 176-241) is run against each the release `check-fw.bin` and our built `firmware-signed.bin`.
```
doc-hex, line 178: - add: "This step verifies the signature of the binary is valid, using either the Coinkite key factory key or the "debug" key zero which is public"

docs/notes-on-repro.md
```diff
@@ -0,0 +1,180 @@
...
+endif
+```
+
+To summarize `check-repro`:
+
+- At the final `check-repro` step, we have a locally built `firmware-signed.bin` and we want to check that it matches 
+the binary release provided by Coldcard.
+- An identical checksum match will not be possible as is, since there is signature data embedded into into the binary, which must be removed.
+- The specific release of the version that is being built is fetched, and placed it under /tmp/checkout/firmware/releases/*.dfu
+- `split` (cli/signit.py: Line 153-175) is run against the release `*.dfu` resulting in a `check-fw.bin` and `check-bootrom.bin`
```
doc-hex, line 177: - add: "This splits the DFU file into the two parts it contains: the main firmware (COLDCARD application) and the boot loader code."

docs/notes-on-repro.md
```diff
@@ -0,0 +1,180 @@
...
+- `check` (cli/signit.py: Line 176-241) is run against each the release `check-fw.bin` and our built `firmware-signed.bin`.
+- a hexdump is taken of each the release `check-fw.bin` and our built `firmware-signed.bin` piped through $TRIM_SIG which removes the signature bytes and subsitutes it with a common string.
```
doc-hex, line 179: Might add here that the signature is 64 bytes, and that is all which is masked out.

docs/notes-on-repro.md
doc-hex: - there are probably other DFU tools that could be used to implement this "split" .. might list some here

i5hi:
@doc-hex can you give me a few notes on the bootrom stuff.

I don't fully understand why its being removed, and how to verify it.

doc-hex:
- bootrom area on Coldcard is not field-upgradable. Set in factory, unchangable.
- the signature on our "firmware" covers the part of flash that can be upgraded
- we store both binaries in the DFU file as a time-saver in the factory where they get both loaded the first time (bootloader and main code)
- there are other ways to verify what bootrom that you have on your Coldcard (but that has nothing to do with build reproducibility)
- the red/green light (checksum check handled by secure element) covers both the bootrom and main firmware (but that doesn't have anything to do wth repro build)

PR #158, opened 2022-10-17
Fix USB framing bug

ghost:
There is a problem with USB framing when handling errors.

1. Inner framing errors weren't floated up for the outer handler to handle so they were handled as generic exceptions.
2. `framing_error` wasn't setting the final (`0x80`) flag on the first byte and that would confuse the other end of the wire and cause it to expect more packets.

The `framing_error` function was rewritten to be more sensible and readable.

If this is accepted, I'll open another PR to port this to the v4 firmware.

### DIFF
--- a/shared/usb.py
+++ b/shared/usb.py
@@ -238,6 +238,9 @@ class USBHandler:
                     # prefer to catch at higher layers, but sometimes can't
                     resp = b'err_Out of RAM'
                     msg_len = 0
+                except FramingError as exc:
+                    # let the outer handler take care of framing errors
+                    raise exc
                 except Exception as exc:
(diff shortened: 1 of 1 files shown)


Terry330:
Want to sell out my coin

Terry330:
Great

Kenny7213:
Can I please have someone contact me to possibly answer a few general
questions I have? I would honestly appreciate you greatly if anyone would
be willing. I have about 95k in tron/usdt but I can't acc3ss the wallet and
its beyond frustrating. I would happily compensate for any progress made
directly from your assistance.

Terry330:
Contact me directly

PR #161, opened 2022-11-30
feat(Advanced, Export Wallet): add Lily Wallet option

KayBeSee:
adds Lily Wallet to Export Wallet menu

This function is essentially the same as generic export but labeled to be more intuitive for less technical users.

### DIFF
--- a/shared/actions.py
+++ b/shared/actions.py
@@ -1145,7 +1145,7 @@ Works for both single signature and multisig wallets. \
 
     # no choices to be made, just do it.
     import export
-    await export.make_json_wallet('Generic Export',
+    await export.make_json_wallet('Lily Wallet',
                                     lambda: export.generate_generic_export(account_num),
                                     'lily-wallet-export.json')
 
--- a/shared/actions.py
+++ b/shared/actions.py
@@ -1131,6 +1131,23 @@ single-signer UTXO associated with this Coldcard.''' + SENSITIVE_NOT_SECRET) != 
                                     lambda: export.generate_generic_export(account_num),
                                     'coldcard-export.json')
 
+async def lily_skeleton(*A):
+    # make a single JSON file with basically all useful XPUB's in it.
+    # identical to generic_skeleton but with different story and filename.
+
+    if await ux_show_story('''\
+This saves a file onto MicroSD card to use with Lily Wallet. \
+Works for both single signature and multisig wallets. \
+''' + SENSITIVE_NOT_SECRET) != 'y':
+        return
+
+    account_num = await ux_enter_number('Account Number:', 9999)
+
+    # no choices to be made, just do it.
+    import export
+    await export.make_json_wallet('Generic Export',
+                                    lambda: export.generate_generic_export(account_num),
+                                    'lily-wallet-export.json')
 
 async def wasabi_skeleton(*A):
     # save xpub, and some other public details into a file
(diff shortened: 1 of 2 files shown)


shared/actions.py
```diff
@@ -1131,6 +1131,23 @@ async def generic_skeleton(*A):
...
+This saves a file onto MicroSD card to use with Lily Wallet. \
+Works for both single signature and multisig wallets. \
+''' + SENSITIVE_NOT_SECRET) != 'y':
+        return
+
+    account_num = await ux_enter_number('Account Number:', 9999)
+
+    # no choices to be made, just do it.
+    import export
+    await export.make_json_wallet('Generic Export',
```
scgbckbone, line 1148: `Generic Export` --> `Lily Wallet`

scgbckbone (commented):
Not against this, but wouldn't it be easier to navigate users to (in some of your tutorials) to `Generic Export` instead? As this is just pure duplication...

KayBeSee:
I get what you're saying, but I think being explicit about it inspires more confidence and removes ambiguity for the user as to what option to select.

For example, I could see them start scratching their head choosing between Generic JSON and the Export XPUB option. And some users might assume the wallet isn't supported since other wallets are explicitly listed in that menu.

nvk:
if we are going to show many vendors. I think we need to have the generics at the top. Then list the many vendors at the bottom. The main issue is that the vendor list will only grow and most will be the same spec as the generic.

kevin, let me take some time to do some thinking with the team on how to scale/handle this.

KayBeSee:
No worries, I appreciate the consideration.

I think you're right that the generics should be on the top of the menu followed by specific wallet vendors. 

An adjacent thought I had this morning is whether `Generic JSON` should be named something like `Universal Export (JSON)` or `Master Export (JSON)` since it contains all of the relevant xpubs/xfps/etc. Then maybe have a plan txt version too (if people aren't about JSON).

Finally, I think the Export XPUB option is kind of confusing since it acts differently than the other options (doesn't save to file). I would consider renaming that to "Show XPUB" since nothing is being exported, but I think the behavior is different with an NFC device? Sorry for dropping a bunch of stuff on you like that, just trying to provide helpful feedback.

nvk:
> No worries, I appreciate the consideration.
> 
> I agree that the generics should be on the top of the menu followed by specific wallet vendors.

The menu topology still needs more thinking as drilling down also sucks. Will do some thinking.

> An adjacent thought I had this morning is whether `Generic JSON` should be named something like `Universal Export (JSON)` or `Master Export (JSON)` since it contains all of the relevant xpubs/xfps/etc. Then maybe have a plan txt version too (if people aren't about JSON).
>

I really like this nomenclature. `Generic` is cheap sounding.
 
> Finally, I think the Export XPUB option is kind of confusing since it acts differently than the other options (doesn't save to file). I would consider renaming that to "Show XPUB" since nothing is being exported, but I think the behavior is different with an NFC device? Sorry for dropping a bunch of stuff on you like that, just trying to provide helpful feedback.

Yes, there are differences here.

KayBeSee:
> > An adjacent thought I had this morning is whether `Generic JSON` should be named something like `Universal Export (JSON)` or `Master Export (JSON)` since it contains all of the relevant xpubs/xfps/etc. Then maybe have a plan txt version too (if people aren't about JSON).
> 
> I really like this nomenclature. `Generic` is cheap sounding.

_Premium_ Export 😉

PR #166, opened 2022-12-20
Address explorer `Applications` and Single Signature Descriptor Export

scgbckbone:
* add "Applications" menu item to address explorer, where application specific derivations paths/addresses are accessible
* fix "change" bug in address file export in address explorer
* add NFC/VDISK options to `write_text_file` and `make_json_wallet` (all wallet exports covered)
* Samourai POST/PRE-MIX account export shortcut
* generic single sig descriptor export -> supports both new descriptor type <0;1> and old with two descriptors exported (internal,external)
* fix write char by char bug write_text_file
* add descriptors(single sig) and descriptor templates (multisig) to generic export json
* move change address switch in address explorer from key (5) to key (6) as you need (5) to move up
* add tests

### DIFF
--- a/shared/descriptor.py
+++ b/shared/descriptor.py
@@ -4,9 +4,9 @@
 #
 # Based on: https://github.com/bitcoin/bitcoin/blob/master/src/script/descriptor.cpp
 #
-from public_constants import AF_P2SH, AF_P2WSH_P2SH, AF_P2WSH
+from public_constants import AF_P2SH, AF_P2WSH_P2SH, AF_P2WSH, AF_P2WPKH, AF_CLASSIC, AF_P2WPKH_P2SH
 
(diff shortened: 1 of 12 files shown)


shared/descriptor.py
```diff
@@ -169,6 +192,98 @@ def serialize_keys(self, internal=False, int_ext=False):
             result.append(key_str.replace("'", "h"))
         return result
 
+    def _serialize(self, internal=False, int_ext=False) -> str:
+        """Serialize without checksum"""
+        assert len(self.keys) == 1, "Multiple keys for single signature script"
```
doc-hex, line 197: make arg to assert be a comment instead (save bytes)

doc-hex: ,... after I merge

PR #171, opened 2023-01-27
Fix SQL2 ImportError on ubuntu 20.04

AbelLykens:
Tried installing simulator on plain ubuntu 20.04, it gave me `ImportError: could not find any library for SDL2 (PYSDL2_DLL_PATH: unset)`. Installing pysdl-dll fixes this.

### DIFF
--- a/unix/requirements.txt
+++ b/unix/requirements.txt
@@ -4,4 +4,4 @@
 #
 PySDL2
 Pillow
-
+pysdl2-dll


doc-hex:
Thanks for investigating this!

I think this might be better as a comment in the README (there is a linux section) because the the `requirements.txt` is used by all systems, not just Ubuntu.

AbelLykens:
Makes sense -- will do and create a new PR :)

PR #196, opened 2023-03-13
WIP: Write logs to microSD in HSM mode

Sosthene00:
Hi, I've been experimenting a lot recently with the CC, especially with the HSM mode, and it happened a couple times that the CC "freezes", it just stops responding at some point without any logs or indication of what happened on the screen. I'd let it run for days, so when it happened I would usually figure it out many hours after. At least once it crashed while signing as the screen was reading "Signing..."

I had a look at what it logged on the microSD and I didn't found it to be very helpful for my situation, so I tried to modify it to write more logs that could help me if it crashes again. 

I tested my code with the simulator, and it worked, I could register a policy and then it would sign psbts and write logs on the SD card.
But then I tried to flash an actual CC Mk4 with my custom code, and it crashed on startup with the following error
photo_2023-03-13_11-53-07

I tried my code again with the simulator, and it works there. Here's an actual example of the content of a log file I just made like this: 
```
44a119cc5d50ceac Info: Loading the psbt file
44a119cc5d50ceac Info: Beginning formal validation of PSBT
44a119cc5d50ceac Error: FatalPSBTIssueNone of the keys involved in this transaction belong to this Coldcard (need 0F056943, found EF9D9CE6, CE3600EA, FACF6B1F, A3A9EB52, 46D93DA5, 86936C07, D3C3BC8F, 8E236875, 7F7D263A, FCA68DB6, D33C583B, D03AACCA, 5DF7E4B0, FE898C92, 154159B3).

===
```
I open this PR as a WIP because I think there might be demand for that, especially as more people will run the HSM mode, it's totally hacky right now but if you agree I'd gladly put on some more work because I need it. I'd just be happy to have a little help and not brick a couple more CC just to figure it out :smile:

### DIFF
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -445,15 +445,15 @@ class ApproveTransaction(UserAuthorizedAction):
             elif wl >= 2:
                 msg.write('(%d warnings below)\n\n' % wl)
 
-            self.output_summary_text(msg)
-            gc.collect()
+            # self.output_summary_text(msg)
+            # gc.collect()
 
             fee = self.psbt.calculate_fee()
             if fee is not None:
                 msg.write("\nNetwork fee:\n%s %s\n" % self.chain.render_value(fee))
 
             # NEW: show where all the change outputs are going
-            self.output_change_text(msg)
+            # self.output_change_text(msg)
             gc.collect()
 
             if self.psbt.warnings:
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -492,7 +492,7 @@ class ApproveTransaction(UserAuthorizedAction):
                     msg.write("\nPress OK to approve and sign transaction. X to abort.")
                     ch = await ux_show_story(msg, title="OK TO SEND?")
                 else:
-                    ch = await hsm_active.approve_transaction(self.psbt, self.psbt_sha, msg.getvalue())
+                    ch = await hsm_active.approve_transaction(self.psbt, self.psbt_sha, msg.getvalue(), log)
                     dis.progress_bar(1)     # finish the Validating...
 
             except MemoryError:
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -391,6 +391,7 @@ class ApproveTransaction(UserAuthorizedAction):
             policy_hash = b'\x00'*32
             never_log = False
         with AuditLogger('logs', self.psbt_sha, never_log, policy_hash) as log:
+            log.new_psbt()
             log.info("Loading the psbt file")
             try:
                 with SFFile(TXN_INPUT_OFFSET, length=self.psbt_len, message='Reading...') as fd:
@@ -408,7 +409,7 @@ class ApproveTransaction(UserAuthorizedAction):
 
             dis.fullscreen("Validating...")
 
-            log.info("Beginning formal validation of PSBT")
+            log.info("PSBT parsing successful")
             # Do some analysis/ validation
             try:
                 await self.psbt.validate()      # might do UX: accept multisig import
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -17,6 +17,7 @@ from utils import B2A, parse_addr_fmt_str
 from psbt import psbtObject, FatalPSBTIssue, FraudulentChangeOutput
 from exceptions import HSMDenied
 from version import has_psram, has_fatram, MAX_TXN_LEN
+from logging import AuditLogger
 
 # Where in SPI flash/PSRAM the two PSBT files are (in and out)
 TXN_INPUT_OFFSET = 0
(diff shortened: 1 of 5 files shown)


scgbckbone:
you need to add your new module `logging.py` to `shared/manifest.py` (or `shared/manifest_mk4`). Simulator does not need this as it loads contents directly from `shared` dir. When building for real thing all modules have to be included in manifest. Your code is failing in `flow.py:22` where it attempts to import from `hsm.py` module. In `hsm.py` it fails because it cannot find module `logging.py` as it is not included in manifest.

Sosthene00:
Thanks, very helpful! I'll try again and try to come up with something cleaner

Sosthene00:
Now it works on the real thing. Here's a summary of the changes made so far:
* Now `AuditLogger` is instantiated as soon as we start parsing the psbt file and passed to `approve_transaction` as an argument. I originally thought I could keep the `AuditLogger` in the `approve_transaction` part and just add another one higher in `auth.py` as to make the less change possible to the way things were, but it appeared that having 2 loggers inside each other writing to the same card was not possible on the real cc (no issue with the simulator though). 
* Instead of having a `psbt` dir with as many log files as psbt submitted, there's now a `log` dir with one file for each hsm policy ('00'*32 being no hsm mode) logging each signing attempt.
* I assumed it was possible to log to card while not being in hsm mode, but maybe it shouldn't be possible idk.
* I started working on the previous release on top of some unrelated changes I was experimenting with and that were the reason I wanted more detailed logs, so I still need to rebase it on the latest release. 

Please let me know if you think this is going in the right direction.

shared/auth.py
```diff
@@ -379,192 +380,214 @@ async def interact(self):
...
+                # mention warning at top
+                wl= len(self.psbt.warnings)
+                if wl == 1:
+                    msg.write('(1 warning below)\n\n')
+                elif wl >= 2:
+                    msg.write('(%d warnings below)\n\n' % wl)
 
-            self.output_summary_text(msg)
-            gc.collect()
+                # self.output_summary_text(msg)
```
scgbckbone, line 471: Commenting out `output_summary_text(msg)` (and `output_change_text(msg)` below) removes all destination addresses/amounts and the change summary from the "OK TO SEND?" approval screen — and from the story passed to `hsm_active.approve_transaction`. The user/HSM is asked to approve a transaction without seeing where the funds go. This is blind signing and defeats the core purpose of the device.

shared/hsm.py
```diff
@@ -855,95 +807,91 @@ def _new_local_code(self):
...
+            # do this super early so always cleared even if other issues
+            local_ok = self.consume_local_code(psbt_sha)
+
+            if not self.rules:
+                raise ValueError("no txn signing allowed")
+
+            # reject anything with warning, probably
+            if psbt.warnings:
+                # if self.warnings_ok:
+                log.info("Txn has warnings, but policy is to accept anyway.")
```
scgbckbone, line 838: The `else: raise ValueError("has %d warning(s)")` branch is commented out, so an HSM policy with `warnings_ok` unset/false now approves transactions that carry warnings anyway. Worse, the log line claims "policy is to accept anyway" even when the configured policy is the opposite, making the audit log misleading.

shared/psbt.py
```diff
@@ -1330,18 +1332,18 @@ def consider_inputs(self):
 
             # iff to UTXO is segwit, then check it's value, and also
             # capture that value, since it's supposed to be immutable
-            if inp.is_segwit:
-                history.verify_amount(txi.prevout, inp.amount, i)
+            # if inp.is_segwit:
+            #     history.verify_amount(txi.prevout, inp.amount, i)
```
scgbckbone, line 1336: Disabling `history.verify_amount(...)` plus the `total_in += utxo.nValue` accumulation (above) means input amounts are never checked against previously seen UTXOs and `total_value_in` is always set to None (line 1349). Consequences: `calculate_fee()` always returns None, so the fee is never shown on the approval screen and no fee warning can ever fire.

shared/psbt.py
```diff
@@ -1199,7 +1199,9 @@ def consider_outputs(self):
 
 
         # check fee is reasonable
-        if self.total_value_out == 0:
+        if self.total_value_in is None:
```
scgbckbone, line 1202: This early `return` fires whenever `total_value_in is None` — which is now always, since `consider_inputs()` unconditionally sets it to None. Everything below is dead code: the "Outputs worth more than inputs!" negative-fee check, the `fee_limit` setting enforcement (`FatalPSBTIssue`), the Big Fee warning, and critically `self.consider_dangerous_change(self.my_xfp)` at line 1225 — so the change-output path heuristics never run and fraudulent change outputs are no longer caught here.

shared/auth.py
```diff
@@ -379,192 +380,214 @@ async def interact(self):
...
-            if isinstance(exc, MemoryError):
-                msg = "Transaction is too complex"
-                exc = None
-            else:
-                msg = "PSBT parse failed"
+        if hsm_active:
+            policy_hash = hsm_active.hash()
+            hsm_rules = hsm_active.save()['rules']
+            try:
+                hsm_rules[0]['never_log']
```
scgbckbone, line 387: This never works as intended: `never_log` is a policy-level field of `HSMPolicy` (`self.never_log`), it never appears in a rule's `to_json()` output, so `hsm_rules[0]['never_log']` always raises KeyError and `never_log` is always False — the HSM `never_log` privacy setting is silently ignored and audit logs are written to the SD card regardless. Additionally, if the policy has zero rules, `hsm_rules[0]` raises IndexError (only KeyError is caught), crashing signing instead of reaching the graceful "no txn signing allowed" refusal; and if the key ever did exist, `never_log` would be unbound (NameError). Use `hsm_active.never_log`.

scgbckbone (commented):
KIMI3:
- Transaction approval screen no longer shows destination addresses/amounts or change summary (`output_summary_text`/`output_change_text` commented out) — users and HSM approve transactions blind
- HSM warning policy bypassed: transactions with warnings are approved regardless of the `warnings_ok` setting
- Input value accounting and UTXO amount verification disabled; `total_value_in` is now always None, so the network fee can never be computed or displayed
- `consider_outputs()` early-returns when `total_value_in is None` (now always), skipping the negative-fee check, the `fee_limit` enforcement, and `consider_dangerous_change()` change-fraud protection
- Broken `never_log` detection: HSM `never_log` privacy policy is ignored (audit log written to SD anyway); empty rules list crashes with IndexError

doc-hex:
I consider this PR abandoned. Closing. If the original author wants to rebase and re-open that's great.

PR #198, opened 2023-03-16
SeedXOR for 12 and 18 words mnemonics

scgbckbone:
add ability to split/recover mnemonic for all supported lengths (12,18,24)

### DIFF
--- a/shared/xor_seed.py
+++ b/shared/xor_seed.py
@@ -5,17 +5,20 @@
 # - for secret spliting on paper
 # - all combination of partial XOR seed phrases are working wallets
 #
-import stash, ngu, chains, bip39, random
-from ux import ux_show_story, ux_enter_number, the_ux, ux_confirm, ux_dramatic_pause
+import stash, ngu, bip39, random
+from ux import ux_show_story, the_ux, ux_confirm, ux_dramatic_pause
 from seed import word_quiz, WordNestMenu, set_seed_value
 from glob import settings
 from actions import goto_top_menu
 
-def xor32(*args):
+def xor(*args):
     # bit-wise xor between all args
-    rv = bytearray(32)
+    vlen = len(args[0])
+    # all have to be same length
+    assert all(len(e) == vlen for e in args)
+    rv = bytearray(vlen)
 
-    for i in range(32):
+    for i in range(vlen):
         for a in args:
             rv[i] ^= a[i]
 
@@ -27,7 +30,7 @@ async def xor_split_start(*a):
 Seed XOR Split
 
 This feature splits your BIP-39 seed phrase into multiple parts. \
-Each part is 24 words and looks and functions as a normal BIP-39 wallet.
+Each part looks and functions as a normal BIP-39 wallet.
 
 We recommend spliting into just two parts, but permit up to four.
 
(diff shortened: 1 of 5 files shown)


shared/xor_seed.py
```diff
@@ -5,17 +5,20 @@
...
 from glob import settings
 from actions import goto_top_menu
 
-def xor32(*args):
+def xor(*args):
     # bit-wise xor between all args
-    rv = bytearray(32)
+    vlen = len(args[0])
+    # all have to be same length
+    assert all(len(e) == vlen for e in args)
```
doc-hex, line 18: this is expensive, and an IndexError will happen on line 23 anyway; remove

shared/xor_seed.py
```diff
@@ -66,32 +69,37 @@ async def xor_split_start(*a):
             if sv.mode == 'words':
                 words = bip39.b2a_words(sv.raw).split(' ')
 
-            if not words or len(words) != 24:
-                await ux_show_story("Need 24-seed words for this feature.")
+            if not words:
+                await ux_show_story("Need seed words for this feature.")
```
doc-hex, line 73: verify menu item protected by a predicate, then remove this check

shared/xor_seed.py
```diff
@@ -66,32 +69,37 @@ async def xor_split_start(*a):
...
                 return
 
-            # checksum of target result is useful.
-            chk_word = words[-1]
+            # checksum of target result is useful (only for 24 words).
+            chk_word = None
+            if len(words) == 24:
+                chk_word = words[-1]
```
doc-hex, line 79: this seems useful at any word length

shared/xor_seed.py
```diff
@@ -199,19 +210,34 @@ async def xor_restore_start(*a):
...
-        msg = "Since you have a seed already on this Coldcard, the reconstructed XOR seed will be temporary and not saved. Wipe the seed first if you want to commit the new value into the secure element."
-        if settings.get('words', 24) == 24:
-            msg += '''\n
-Press (1) to include this Coldcard's seed words into the XOR seed set, or OK to continue without.'''
+        msg = ("Since you have a seed already on this Coldcard, the reconstructed XOR seed will be "
+               "temporary and not saved. Wipe the seed first if you want to commit the new value "
+               "into the secure element.")
+        if curr_num_words == desired_num_words:
+            msg += ("\nPress (1) to include this Coldcard's seed words into the XOR seed set, "
+                    "or OK to continue without.")
```
doc-hex, line 240: block '1' below as well

shared/xor_seed.py
scgbckbone: expensive yet needed. `IndexError` not thrown for example if first element (from which we take `vlen`) is smaller then rest. Performance penalty is minimal as tuple cannot have more than 4 items to check.

shared/xor_seed.py
scgbckbone: you're right - I somehow dismissed the idea in the beginning of the task thinking that we do not have enough bits to reconstruct. But the idea is wrong as we have the seed that we're trying to reconstruct at hand. Thanks - fixed

shared/xor_seed.py
scgbckbone: still buggy -> [12,24,12]

we must be strict here

doc-hex: imho a range-checking assertion that isn't as narrow as possible isn't a bug, but fine... keep your version

PR #205, opened 2023-03-31
UX repetition fix

scgbckbone:
Mk3 backport


brianddk:
I'm watching this PR in the hopes of resolving MK3 repro builds.  If anyone could suggest a commit tag to build on, I could share my build results to see if we can get two independent matches.

Any thoughts?

* https://github.com/brianddk/attestation/issues/10
* https://github.com/brianddk/attestation/issues/11

doc-hex:
Standby! We're going to make an Mk3 release later this week, and then it will be easy to repro-build it.

brianddk:
Thx @doc-hex , I verified 4.1.8 looks good

PR #211, opened 2023-05-09
Taproot

scgbckbone:
* taproot single sig keypath spend
* 1st iteration of very limited tapscript multisig. Only tapscript of (tree)depth 0 is allowed (meaning leaf_hashs == merkle_root) - only allowed script is `sortedmulti_a(` (I even kept artificial limit of N=16 from p2sh legacy multisig - can be much higher). Mostly same as normal `sortedmulti(`. Must be registered first. Only difference is that you must have `internal_key` defined ---> `tr(internal_key, TREE(optional))` we only allow exactly 1 `sortedmulti_a(` ---> `tr(internal_key, sortedmulti_a(MofN))`. We allow 1.) fully specified ranged xpubs as internal_key 2.) provably unspendable xonly pubkey from BIP34
* address explorer with support to p2tr addresses (both single and multisig)
* generic export, bitcoin-core export and dump summary now export taproot extended keys (and addresses for summary)
* taproot paper wallets
* more info in `docs/taproot.md`


shared/address_explorer.py
```diff
@@ -27,14 +27,15 @@ def __init__(self, path=None, nl=0):
...
-                    ]
+                MenuItem("m/..", f=self.deeper),
+                MenuItem("m/44'/..", f=self.deeper),
+                MenuItem("m/49'/..", f=self.deeper),
+                MenuItem("m/84'/..", f=self.deeper),
+                MenuItem("m/86'/..", f=self.deeper),
+                MenuItem("m/0/{idx}", menu=self.done),
+                MenuItem("m/{idx}", menu=self.done),
+                MenuItem("m", f=self.done),
+            ]
```
doc-hex, line 38: please restore previous order: it was intentional

doc-hex:
also includes BSMS support

PR #217, opened 2023-05-16
Tapleaf version is only 7 most significant bits

scgbckbone:
* PSBTs do not contain pk parity as it is only needed for witness stack
* bitcoin core correctly sends versions without parity in PSBT
* I have not seen other wallets (mostly because tapscript not implemented) doing it - BUT to be on safe side

### DIFF
--- a/shared/chains.py
+++ b/shared/chains.py
@@ -7,7 +7,8 @@ from uhashlib import sha256
 from ubinascii import hexlify as b2a_hex
 from public_constants import AF_CLASSIC, AF_P2SH, AF_P2WPKH, AF_P2WSH, AF_P2WPKH_P2SH, AF_P2WSH_P2SH, AF_P2TR
 from public_constants import AFC_PUBKEY, AFC_SEGWIT, AFC_BECH32, AFC_SCRIPT
-from serializations import hash160, ser_compact_size, disassemble
+from public_constants import TAPROOT_LEAF_TAPSCRIPT, TAPROOT_LEAF_MASK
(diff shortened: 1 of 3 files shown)


doc-hex:
- feels like it should be an fatal error (bad psbt), if set
- we might sign something differently from another wallet that isn't smart enough to mask out the bit

shared/chains.py
```diff
@@ -37,6 +38,12 @@ def taptweak(internal_key, tweak=None):
     xo_pubkey_tweaked = xo_pubkey.tweak_add(tweak)
     return xo_pubkey_tweaked.to_bytes()
 
+def tapleaf_hash(script, leaf_version=TAPROOT_LEAF_TAPSCRIPT):
```
scgbckbone, line 41: unified this into func - when hashing we only care about version

shared/psbt.py
```diff
@@ -1954,6 +1956,7 @@ def sign_it(self):
                                 if which_key in script:
                                     taproot_script = script
                                     leaf_ver = lv
+                                    assert leaf_ver == TAPROOT_LEAF_TAPSCRIPT, "tapleaf ver"
```
scgbckbone, line 1959: won't sign anything other than version that we know (0xc0)

scgbckbone (commented):
more restrictive

shared/chains.py
scgbckbone: https://github.com/bitcoin/bitcoin/blob/master/src/script/interpreter.cpp#L1929

PR #231, opened 2023-06-21
Initial repro github action

praveenperera:


doc-hex:
I prefer these rebuilds are done by humans (with screen recording on)... just having Github do something in the background doesn't feel satisfying

praveenperera:
Ya fair enough, I didn't mean to open a PR here. But good CI to have just to make sure the repro script is working.

PR #249, opened 2023-08-22
update libngu to master tip

scgbckbone:
* taproot code paths included but NOT used
 * (more) libsecp256k1 context randomization (side-channels)


AngusP:
(maybe related, I had a build fail from pre- https://github.com/bitcoin-core/secp256k1/pull/1009 with the current `libngu` on Mac Sonoma -- though I don't think you're pulling that diff in here?)

PR #262, opened 2023-09-19
remove obsolete mk2/mk3 code paths from firmware

scgbckbone:
Removed old Mk2/Mk3 code that is no longer used.


shared/seed.py
```diff
@@ -637,18 +635,8 @@ def clear_seed():
...
-
-        if version.has_608:
-            # wipe the long secret too
-            nv = bytes(AE_LONG_SECRET_LEN)
-            pa.ls_change(nv)
+    callgate.fast_wipe(True)
+    # NOT REACHED
 
     dis.busy_bar(False)
     dis.fullscreen('Reboot...')
```
doc-hex, line 642: these 2 lines are unreachable

doc-hex:
- `mk-sigheader.py` and related files are a bit messed up? Hard to tell from the diff. 
- all of the changes to `/stm32/` are not helpful since they do not reduce memory size.

shared/seed.py
scgbckbone: fixed in https://github.com/Coldcard/firmware/pull/263

scgbckbone:
> * `mk-sigheader.py` and related files are a bit messed up? Hard to tell from the diff.
> 
>     * all of the changes to `/stm32/` are not helpful since they do not reduce memory size.

reverted whole `stm32` (we can do this in solo PR later)

scgbckbone:
https://github.com/Coldcard/firmware/pull/263

PR #264, opened 2023-09-20
BIP39 passphrase as ephemeral seed; Lock Down Seed for all ephemeral;BIP-39 wallet backup

scgbckbone:
* bip39 passphrase is now internally handled as ephemeral secret
* bip39 passphrases now have its own encrypted settings
* add ability to backup bip39 passphrase wallet
* bip39 passphrases can now be added to ephemeral (word based) secrets
* new feature to make normal main seed(Lock Down Seed) from any currently loaded ephemeral (previously feature called bip39 to bip32 and only implemented for passphrases)
* removed ability to see BIP-39 Passphrase after wallet is active
* showing secrets now also display extended private key for BIP-39 passphrase wallets.

### DIFF
--- a/docs/backup-files.md
+++ b/docs/backup-files.md
@@ -38,6 +38,17 @@ a single file, which is a simple text file and
 easy to read. Before version 4.0.0, this text file was always
 called `ckcc-backup.txt`, but the filename is now picked randomly.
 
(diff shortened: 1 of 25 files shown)


scgbckbone:
re-based on top of updated master

PR #276, opened 2023-10-06
Refactor nvstore  

doc-hex:
- WIP
- planning another round


scgbckbone:
just skimmed --> much better than subclass idea!

I will defer my review till you done as you're "planing another round"

scgbckbone:
this is obsolete now and included here https://github.com/Coldcard/firmware/pull/277

PR #291, opened 2023-11-22
IOError is deprecated, use OSError

scgbckbone:


doc-hex:
IMHO remove from ChangeLog rather than rebase

scgbckbone:
make sense - too internal anyways - changelog line removed

PR #292, opened 2023-11-22
IOError is deprecated, use OSError (EDGE)

scgbckbone:


scgbckbone:
already in `edge` HEAD

PR #297, opened 2023-12-07
bugfix: do not allow to import master seed as temporary

scgbckbone:


shared/seed.py
```diff
@@ -471,13 +471,13 @@ async def set_ephemeral_seed(encoded, chain=None, summarize_ux=True, bip39pw='',
...
 
-    applied = pa.tmp_secret(encoded, chain=chain, bip39pw=bip39pw)
+    applied, err_msg = pa.tmp_secret(encoded, chain=chain, bip39pw=bip39pw)
     dis.progress_bar_show(1)
     xfp = settings.get("xfp", None)
     if xfp:
         xfp = "[" + xfp2str(xfp) + "]"
     if not applied:
-        await ux_show_story(title=xfp, msg="Temporary master key already in use.")
+        await ux_show_story(title="FAILED", msg=err_msg)
```
doc-hex, line 480: more if to after line 475

scgbckbone: yes - fixed https://github.com/Coldcard/firmware/pull/297/commits/0bbbd579dc9714fde93c126a3160e9759123e890

PR #303, opened 2023-12-14
pwsave menu UX rework; do not allow empty bip39 passphrase

scgbckbone:
* do NOT allow to `APPLY` empty bip39 passphrase. In case of empty string `APPLY` is NOOP
* allow user to delete entries from `.tmp.tmp`. Removing last saved passphrase also removes file itself
* do not store duplicate entries into `.tmp.tmp`


scgbckbone:
rebased

PR #314, opened 2023-12-30
remove ability to use `Passphrase` if already in use

scgbckbone:
* this was kept as legacy UI/UX compatibility when passphrase was migrated from global var tmp value to classic way of doing tmp seed.
* main reason for this feature to exist in a first place was inability to switch from tmp to master seed without reboot- now have `Restore Master`

### DIFF
--- a/releases/ChangeLog.md
+++ b/releases/ChangeLog.md
@@ -1,3 +1,9 @@
+## 5.2.3 - 2024-XX-XX
+
+- Change: `Passphrase` menu item is no longer offered if BIP39 passphrase
+  already in use. Use `Restore Master` with ability to keep or purge current
+  passphrase wallet settings.
+
 ## 5.2.2 - 2023-12-21
 
 - Bugfix: Re-enable `Lock Down Seed` feature which was disabled by accident
--- a/shared/flow.py
+++ b/shared/flow.py
@@ -51,10 +51,8 @@ def se2_and_real_secret():
     from pincodes import pa
     return (not pa.is_secret_blank()) and (not pa.tmp_value)
 
-def bip39_passphrase_active():
-    import stash
-    return settings.get('words', True) \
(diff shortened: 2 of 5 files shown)


MathieuBerthellemy:
Hello, I'm very recent user of the coldcard (5.2.0), In the first place I came with the same though, once the passphrase is set and the "Restore Master" available why still having the "passphrase" option ? 

After try and fail I figure it out that if you redefine the passphrase in the coldcard it'll just replace the current passphrase with a new one, this scenario can be simplified as you suggest by the sequence -> passphrase -> Restore Master -> passphrase

But a new usecase cross my mind, wouldn't be wonderful to be able to apply a passphrase on top of the previous passphrase ? I would act like a "26th word", if this idea makes sense, that what "passphrase" should do if a passphrase is already setted.

scgbckbone:
> Hello, I'm very recent user of the coldcard (5.2.0), In the first place I came with the same though, once the passphrase is set and the "Restore Master" available why still having the "passphrase" option ?

As you correctly pointed out in next paragraph - it just adds new passphrase on the top of the master seed, completely ignoring current set passphrase. This was needed historically as `Restore Master` option is very recent. 

> But a new usecase cross my mind, wouldn't be wonderful to be able to apply a passphrase on top of the previous passphrase ? I would act like a "26th word", if this idea makes sense, that what "passphrase" should do if a passphrase is already setted.

Not possible as BIP39 Passphrase wallets are extended key based (tprv, tprv) and you can only add passphrase to word based wallets.

MathieuBerthellemy:
Gotcha, thank you to have pointed out this behavior. in this scenario it’s desirable to remove “passphrase” entry if passphrase is already setted

scgbckbone:
closing here as this PR will be part of Q merge to master

PR #319, opened 2024-01-18
Bugfix: pwsave froze (edge)

scgbckbone:
* already in `master`
* improved exception handling (will port back to master)


doc-hex:
yes good stuff

PR #325, opened 2024-02-05
bugfix: multisig ascii input validation

scgbckbone:
* only allow printable ascii characters `range(32, 127)` as multisig wallet name
* push `ckcc-protocol` to current master HEAD


scgbckbone:
- [ ] add `limitations.md`

PR #327, opened 2024-02-06
always use SettingsObject class for master data

scgbckbone:
* this is NOT a bug fix as using instance `settings` work properly on real HW which requires a reboot
* unification of access via class instance
* fixes `test_bip39pw.py` on simulator


MasterOwnerETcorps (commented):
¶`<--'"<!m>={" create&merge:"} ["new pull request;"] 'On', ("#/.mpret").-->`

PR #329, opened 2024-02-18
(unreleased) `5.2.3` master port (edge)

scgbckbone:


scgbckbone:
`edge` branch was re-built [here](https://github.com/coinkite/afirmware/pull/237) with all above included

PR #332, opened 2024-02-22
Q Firmware

doc-hex:
This PR will hold the Q related firmware changes until they are merged.

Coldcard Q is not yet shipping, but you can reserve [your unit at our store.coinkite.com](https://store.coinkite.com/store/category/reservations).  

Built and signed binaries for the lucky few who have test Q hardware can be found here: <https://coldcard.com/downloads>

This branch will build both Mk4 and Q binaries with the same features (subject to hardware differences) but we have not yet tested the many changes on Mk4. 

- [Q Release notes](https://github.com/Coldcard/firmware/blob/q1staging/releases/QChangeLog.md)
- [Q vs Mk4 Hardware comparison](https://coldcard.com/docs/coldcard-q/)


doc-hex:
[Q Release notes](https://github.com/Coldcard/firmware/blob/q1staging/releases/QChangeLog.md)

shared/manifest_q1.py
```diff
@@ -0,0 +1,29 @@
...
+	'gpu.py',
+	'vdisk.py',
+	'nfc.py',
+	'ndef.py',
+	'trick_pins.py',
+	'ux_q1.py',
+	'battery.py',
+	'notes.py',
+	'calc.py',
+], opt=0)
```
AngusP, line 21: Noticed this is missing the files needed for HSM mode, from reading the other code I don't *think* removing that feature from the Q1 was intentional?

```diff
--- manifest_q1.py
+++ manifest_q1.py
@@ -15,6 +15,9 @@
        'ndef.py',
        'trick_pins.py',
        'ux_q1.py',
+       'hsm.py',
+       'hsm_ux.py',
+       'users.py',
        'battery.py',
        'notes.py',
        'calc.py',
```

**Mk4:**

```python
freeze_as_mpy('', [
	'ssd1306.py',
	'mempad.py',
	'psram.py',
	'mk4.py',
	'vdisk.py',
	'nfc.py',
	'ndef.py',
	'trick_pins.py',
	'ux_mk4.py',
	'hsm.py',
	'hsm_ux.py',
	'users.py',
], opt=0)
```

doc-hex: We wanted the flash space for the new larger font and other good stuff. In HSM mode, the Mk4 can do everything the Q can do, since the QR scanner and other improvements are primarily UI that would be wasted in unattended mode.

AngusP: Ah, that makes sense -- is there not enough flash space at all to include it, or would it eat settings space or something? Slgihtly sad but I'll live 😆 

Might be handy to note that it's only Mk4 (and Mk3?) for HSM on the HSM docs https://coldcard.com/docs/hsm/ and Q1 webpage https://coldcard.com/q , I couldn't find a note anywhere other than digging in to the code here (though yes it's a pretty advanced topic so that could be TMI/clutter)

doc-hex: You are right, we need to document this. I'll follow-up on that.

PR #336, opened 2024-03-22
bugfix: backward compatible fix of AFC_BECH32M

scgbckbone:


shared/address_explorer.py
```diff
@@ -344,7 +344,8 @@ def make_msg(change=0):
                     continue
 
                 from ux import show_qr_codes
-                await show_qr_codes(addrs, bool(addr_fmt & AFC_BECH32), start)
+                is_alnum = bool((addr_fmt & AFC_BECH32) or (addr_fmt & AFC_BECH32M))
```
doc-hex, line 347: `bool(addr_fmt & (AFC_BECH32|AFC_BECH32M))`

scgbckbone: right

PR #337, opened 2024-03-27
bugfix: b39pw on tmp when SE secret not set

scgbckbone:
* would produce a Yikes
* properly distinguish between having SE secret and having tmp secret


scgbckbone:
already in Q

PR #338, opened 2024-04-06
Update typo in paper.py

gzuuus:
I guess this is a typo, isn't it?


doc-hex:
I think dice is the plural of die.

gzuuus:
Yes, sorry for the confusion, it is true that it seems that both 'dice' and 'die' are accepted with singular of die nowadays.

PR #366, opened 2024-05-30
tx output explorer

scgbckbone:
* add ability to "explore" all tx outputs (useful txs with out > 10)
*  fix `fake_ms_txn` and add psbtV2 to fixture
* add ability to add multiple `op_return`s to `fake_txn` fixture

### DIFF
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -21,7 +21,7 @@ from psbt import psbtObject, FatalPSBTIssue, FraudulentChangeOutput
 from files import CardSlot
 from exceptions import HSMDenied
 from version import MAX_TXN_LEN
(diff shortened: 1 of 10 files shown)


shared/auth.py
```diff
@@ -836,6 +846,52 @@ async def interact(self):
...
+        from glob import dis
+        start = 0
+        n = 10
+
+        def make_msg(start, n):
+            dis.fullscreen('Wait...')
+            rv = ""
+            end = start + n
+            if end > self.psbt.num_outputs:
+                end = self.psbt.num_outputs
```
doc-hex, line 859: use min() here

PR #368, opened 2024-06-03
NFC Push Tx feature

doc-hex:
This feature allows single-tap broadcast of the freshly-signed transaction.

Once enabled with a URL, the COLDCARD will show the NFC animation
after signing the transaction. When the user taps their phone, the
phone will see an NFC tag with URL inside. That URL contains the
signed transaction ready to go, and once opening in the mobile
browser, that URL will load. The landing page will connect to a
Bitcoin node (or similar) and send the transaction on the public
Bitcoin network.

### DIFF
--- a/docs/nfc-pushtx.md
+++ b/docs/nfc-pushtx.md
@@ -1,0 +1,76 @@
+# NFC Push Tx
+
+This feature allows single-tap broadcast of the freshly-signed transaction.
+
+Once enabled with a URL, the COLDCARD will show the NFC animation
+after signing the transaction. When the user taps their phone, the
+phone will see an NFC tag with URL inside. That URL contains the
+signed transaction ready to go, and once opening in the mobile
+browser, that URL will load. The landing page will connect to a
+Bitcoin node (or similar) and send the transaction on the public
+Bitcoin network.
+
+This feature is available on Q and Mk4 and requires NFC to be enabled.
+See `Advanced/Tools > NFC Push Tx`
+
+## Protocol Spec
+
+The COLDCARD needs a URL prefix. To that it appends some values:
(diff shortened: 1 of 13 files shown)


docs/nfc-pushtx.md
```diff
@@ -0,0 +1,76 @@
+# NFC Push Tx
+
+This feature allows single-tap broadcast of the freshly-signed transaction.
+
```
nvk, line 4: Add here:

PSBT ==[SD|QR|NFC]==> COLDCARD signed TXN ==[NFC tap]==> Phone Browser ==> Server TXN Broadcast

shared/auth.py
```diff
@@ -806,9 +806,20 @@ async def interact(self):
         except BaseException as exc:
             return await self.failure("PSBT output failed", exc)
 
-        from glob import NFC
+        from glob import NFC, settings
 
         if self.do_finalize and txid and not hsm_active:
+
+            # if NFC PushTx is enabled, do that w/o questions.
```
scgbckbone, line 813: meh for starting without confirmation of action. Would be much better inside the loop on line 826. This way users can export multiple times on possible failure

scgbckbone (commented):
should also have an `NFC Tools` entry to read signed txn from microSD and share PushTx url

shared/auth.py
doc-hex: disagree. they enabled it so they want it. already loops for multiple attempts when the tap fails / phone craps out, etc.

doc-hex:
> should also have an `NFC Tools` entry to read signed txn from microSD and share PushTx url

yes, good idea ... will add

PR #369, opened 2024-06-04
add shortcut to PSBT signing in `NFC tools`

scgbckbone:
It is extremely annoying to me that I have to remove SD card every-time I want to use NFC signing (SD card contains PSBTs). `Sign PSBT` menu item in `NFC Tools` menu solves this as it provides direct access to NFC signing.  Just 2 clicks on Q --> NFC button + ENTER

* improve exception messages for NFC Tools menu items (added `problem_file_line`)
* wrap address verification to try/except to prevent yikes


RandyMcMillan:
Concept ACK

The "3 click rule" is still a useful guideline in UI/UX design for HWW. 

REF: https://en.wikipedia.org/wiki/Three-click_rule

PR #392, opened 2024-07-11
Allow unsorted multi to be handled as miniscript

reardencode:
I did some testing and found that
`wsh(and_b(multi(2,[xfp/84h/0h/0h]xpub.../1/*,[xfp]xpub.../1/*,[xfp]xpub.../1/*),a:older(1000)))` works, so there doesn't seem to be an internal reason not to also support top level unsorted multi.

### DIFF
--- a/shared/miniscript.py
+++ b/shared/miniscript.py
@@ -328,7 +328,7 @@ class MiniScriptWallet(BaseStorageWallet):
         else:
             name = to_ascii_printable(name)
             desc_obj = Descriptor.from_string(config.strip())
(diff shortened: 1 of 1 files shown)


scgbckbone:
Thanks @reardencode ! You're right, in edge firmware we can handle `multi()` anywhere, no problem, and restriction is completely artificial. In normal firmware, some more tweaking needs to be done to allow non-BIP-67 multisig. Good news is that `multi()` support is coming to normal firmware - very soon. 

After support is merged in master it will be ported to edge too.

scgbckbone:
https://github.com/Coldcard/firmware/pull/408

PR #395, opened 2024-07-18
feat: add Fully Noded, compatible with FN v1.2.0

Fonta1n3:
FN v1.2.0 offers full multisig and single sig support for Coldcard Q1 bbqr wallet and psbt's flow. This PR adds Fully Noded to the list of wallet export options.

coldcard_fullynoded


doc-hex:
Awesome, thanks!

PR #396, opened 2024-07-25
Added necessary step to install instructions

andreasgriffin:
- Added necessary step to install instructions
- added numpy as python requirement

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ cd firmware
 git apply unix/linux_addr.patch
 
 #  * below is needed for ubuntu 24.04
-pushd ../external/micropython
+pushd external/micropython
 git apply ../../ubuntu24_mpy.patch
 popd
 #  * 


doc-hex:
Looks fine, can I merge? It's still marked as draft.

unix/requirements.txt
```diff
@@ -4,4 +4,4 @@
 #
 PySDL2
 Pillow
-
```
scgbckbone, line 7: maybe you had an install failure? `numpy` is required by `zbar-py` which is listed in `testing/requirements.txt`

README.md
```diff
@@ -185,6 +185,8 @@ cd firmware
 # Apply address patch
 git apply unix/linux_addr.patch
 
+cd unix
```
scgbckbone, line 188: remove and instead change below to `pushd external/micropythohn`. Changing dir to unix here would break cmds below

scgbckbone (changes requested):

unix/requirements.txt
andreasgriffin: probably a mistake from the 
cd unix

Testing now.

andreasgriffin:
Now the install instructions work flawlessly :-)

scgbckbone (approved):
good catch - thanks @andreasgriffin !

PR #399, opened 2024-07-26
bugfix: do not allow to import duplicate ms wallet with just shuffled keys

scgbckbone:
Currently possible to import both `sortedmulti(2,A,B)` and `sortedmulti(2,B,A)`. Treat as duplicates and sort xpubs before comparison.


scgbckbone:
rebased

PR #405, opened 2024-08-07
WIP: fair coin seed generation

mcudev:
This PR is a work in progress. I am curious, is there interest in accepting a seed generation via a fair coin process (before I work on the device specific GUI code)? The process is conceptually what I've implemented in this PR's python code. The functionality would be available on device in the menu structure as indicated in this PR. I discussed adding something similar to the Trezor devices a little over a year ago [10][11][12][13], but have not received feedback.

**Motivation**

Basically, I want a stupidly simple workflow to guide the user through:

1. generating unique, unpredictable, **unbiased**, untamperable entropy, uniformly at random
2. capturing that entropy into their trusted device
3. verifying that the input entropy accurately maps to the generated BIP39 mnemonic sentence

Entropy is a measure of unpredictability. A perfectly biased coin/die is totally predictable, and provides 0 entropy. Thus:
A) 1 toss of a coin, having unknown bias, can provide from 0 to 1 bit of entropy.
B) 1 roll of a D6, having unknown bias, can generate between 0 and log2(6) = ~2.585 bits of entropy.
C) Tossing/rolling a set of coins/dice having unknown bias does not mitigate vulnerability to bias. **Not knowing the bias is not the same as there not being bias.**

For our purposes, bias is bad. We want to eliminate bias in the generated entropy. The problem is that the bias of users' coins/dice is not known and can not be appraised.
To remove bias from coin trials, I prefer a simple, repeatable and verifiable von Neumann process, see references below [1][2][3][4].

The von Neumann process implemented here gives us a fair coin from a likely imperfect coin. That is, the coin itself yields a result with unbiased, equal probability. However, the process of how the coin is tossed still should be as unpredictable as possible in order to generate unpredictable trial observations, and thus good result "entropy". That is, we should not encourage tossing devices that, for example, replicate closely the energy imparted to, and launch angle of the coin.

The math that gives us equal probability from the 2-trial pairs of tosses:
P(heads) = H
P(tails) = 1 - H = T
Total probability from 2 trials = HH + TH + HT + TT

We don't know what the probabilities of H and T actually are (that is, we don't know how fair/unfair the coin actually is).
We do know that TH and HT are equal (commutative property of multiplication) and that HH and TT are likely unequal (unless the coin is already perfectly fair). Thus, we discard in the 2 cases having likely unequal probability (HH and TT).

The goal is to allow the user to keep track as the process progresses, for verification purposes, to assure them that the input is exactly what is being used to generate the words. As they go, they want to know whether a 0, 1, or nothing was generated. Then they want to see the 11 bits + the word that the bits map to for wordlist verification. It's a high work, high verification, high assurance process. User efficiency with respect to wall time is not the goal of this process.

An interesting part of this implementation is that every time a valid bit is generated, we show the current set of 11 bits for user verification and progress. After the 11th bit of each word is input, we display the generated word. The user can keep track of the word and verify that it was selected according to the input entropy by looking the word up in a BIP39 wordlist such as [5] (we could show the binary and hex to match the wordlist).
This supports the request made in [9] for verifiability of the mnemonic sentence against the input entropy.

**Note**

Regarding SHA-256 for randomness extraction from your D6 dice rolls. I assume that decision was made in order to more efficiently extract closer to ~2.585 bits per roll.
That process does not account for dice bias (which likely increases when using low quality dice). Also, by using SHA-256 to extract randomness, I believe that you're reducing the strength of the extracted entropy by some unkown amount. That is, I believe that process is not reliably getting "full entropy" in the output [6][7][8]. To reliably achieve full entropy out of SHA-256, that process would have to input twice the output block size, which would mean 512 bits of entropy.

References
[1] https://en.wikipedia.org/wiki/Fair_coin#Fair_results_from_a_biased_coin
[2] https://en.wikipedia.org/wiki/Randomness_extractor#Von_Neumann_extractor
[3] https://gist.github.com/atoponce/c594f6463cdf067ea8cabd899939c24f#file-inequality-2-md
[4] https://crypto.stackexchange.com/a/52682
[5] https://github.com/Coldcard/wordlist-paper/blob/master/wordlist.pdf
[6] https://crypto.stackexchange.com/a/75706
[7] https://crypto.stackexchange.com/a/58168
[8] Section 6.4.2.2 https://csrc.nist.gov/csrc/media/publications/sp/800-90b/draft/documents/draft-sp800-90b.pdf
[9] https://github.com/trezor/trezor-firmware/issues/1293#issuecomment-701921550
[10] https://github.com/trezor/trezor-firmware/issues/23#issuecomment-1535660039
[11] https://github.com/trezor/trezor-firmware/issues/23#issuecomment-1559266504
[12] https://github.com/trezor/trezor-firmware/issues/23#issuecomment-1577774027
[13] https://github.com/trezor/trezor-firmware/issues/23#issuecomment-1577783411


mcudev:
closing. good luck.

PR #408, opened 2024-08-13
Opt-in allow `multi(...)`

scgbckbone:
* allow unsorted multisigs that do not follow [BIP-67](https://github.com/bitcoin/bips/blob/master/bip-0067.mediawiki)
* opt-in: must be activated in `Settings->Multisig Wallets->Legacy Multisig`
* only allow to import/export via descriptors ( COLDCARD import/export format is BIP-67 only)
* `multi()` multisig wallets are backwards incompatible - cannot be imported or restored from backup on older firmware versions 
* allow JSON wrapped descriptor imports
* do not require descriptor checksum on import anymore


releases/Next-ChangeLog.md
```diff
@@ -4,11 +4,16 @@ This lists the new changes that have not yet been published in a normal release.
 
 # Shared Improvements - Both Mk4 and Q
 
+- New Feature: Opt-in support for `multi(...)` - unsorted multisig. Enable here `Settings->Multisig Wallets->Legacy Multisig`
```
reardencode, line 7: It'd be great if this didn't say "legacy" since BIP67 is now obsoleted by descriptors and the only way to use this is with descriptors.

Perhaps "Unsorted Multisig" or "Descriptor Multisig" or "multi Descriptor"

scgbckbone: agreed https://github.com/Coldcard/firmware/pull/408/commits/39eef7234ef4c722d315dbed80173d0653c3cc54

reardencode: TYVM! (that commit doesn't appear in this PR, not sure if it is expected to)

scgbckbone: strange, when I click on the link above, I can see the note that the commit is from fork and does not belong to this repo (which is correct). When I go via PR->commits->39eef72 I do not see the note. 

As long as you can see the diff (?), It's OK.

PR #425, opened 2024-09-29
Fix grammar error

henrialb:
This PR fixes a small grammar error I found in the Settings menu, inside "Menu Wrapping".


doc-hex:
Good catch, thanks!

PR #431, opened 2024-11-20
Edge Nov2024

scgbckbone:
* contains critical BUGFIX wrt to policy filling https://github.com/Coldcard/firmware/commit/1a4116427123d4eb3741e43419bc038a43452928
    * creating descriptor from serialized storage policy contains issue, causing policy to be incorrectly filled which causes wallet to stop working right after successful import/enroll (user is unable to sign txs with the wallet) for complex policies that contain same origin keys with just different subderivation.
    * bug introduced in version `6.2.2X`

* ports from `5.4.0mk4` and `1.3.0Q`


scgbckbone (commented):
expanding multisig tests https://github.com/Coldcard/firmware/pull/431/commits/49de639bacd9bd0708b4dc6c4dbcd9024d43009a

PR #437, opened 2024-12-15
bugfix: single key miniscript wallets

scgbckbone:


scgbckbone:
* UX fix wrt checkmark https://github.com/Coldcard/firmware/pull/437/commits/8d5debfb5bc0cd71ba17b0f5142e399c3d1543d1

PR #440, opened 2025-01-01
add missing word to seed xor docs

tyler-dot-earth:
i added what i'm guessing is the missing word to the xor seed docs.

i also removed a double space following a period - partly to keep the line length to 80, but it's also more consistent with the rest of the file.


doc-hex:
Yes, thanks!

PR #441, opened 2025-01-02
deltamode & xor_seed

dmonakhov:
Die rather than give up our secrets

### DIFF
--- a/shared/xor_seed.py
+++ b/shared/xor_seed.py
@@ -70,6 +70,11 @@ Otherwise, press {ok} to continue.'''.format(n=num_parts, ok=OK), escape='2')
     raw_secret = bytes(32)
     try:
         with stash.SensitiveValues() as sv:
+            if sv.deltamode:
+                # die rather than give up our secrets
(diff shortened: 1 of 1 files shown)


dmonakhov:
Use same approach as https://github.com/Coldcard/firmware/commit/8f86ed1c0ed29d0e5e7891dee2bc26d6196a26d9

shared/xor_seed.py
```diff
@@ -31,6 +31,11 @@ def xor(*args):
 
 async def xor_split_start(*a):
 
+    from pincodes import pa
```
scgbckbone, line 34: move this below line 77 and use `SensitiveValues` object `deltamode` attribute

```
        if sv.deltamode:
            import callgate
            callgate.fast_wipe()
```

dmitry-monakhov: Done,
Plus also add a check for SeedXOR recover, since it is not a good idea to offer master/seedvault secrets for xor recovery in deltamode.

PR #442, opened 2025-01-07
testing: cope with buggy core 28.0 behavior wrt labels in importdescriptors

scgbckbone:
* remove any labels from ranged descriptors


doc-hex:
sure

PR #514, opened 2025-04-13
Fix: Correct PSBT finalization for multi-step signing scenarios

uebber:
Problem:
The previous `finalize` method incorrectly assumed that all signatures required for an input would be present in the `added_sigs` dictionary, which only holds signatures generated during the *immediately preceding* `sign_it` call.

This caused failures in scenarios involving multiple signing steps for the same PSBT, such as:
1.  Multi-signature wallets where different signers sign sequentially.
2.  Single-signature wallets using multiple BIP39 passphrases/logins on the same device, where inputs belonging to different internal wallets are signed in separate `sign_it` calls.

In these cases, signatures from earlier steps are stored in `part_sigs`. The `finalize` method would fail with an assertion error (`'No signature on input #%d'` or `'Missing signature for input #%d'`) when processing an input whose signature resided only in `part_sigs`.

Changes:
1.  **Completeness Check:** Modified the initial assertion within the input loop of `finalize`.
    *   For single-signature inputs, it now checks `len(inp.part_sigs) + len(inp.added_sigs) >= 1` to ensure at least one signature exists across both dictionaries.
    *   For multi-signature inputs, the existing check using `self.multi_input_complete(inp)` was already correct as the helper function properly considers both `part_sigs` and `added_sigs` against the required threshold `M`. An explicit check for `self.active_multisig` context was added before calling this helper.
2.  **Signature Retrieval (Single-Sig):** Updated the logic for constructing the final `scriptSig` (non-SegWit) and witness stack (SegWit) for *single-signature* inputs.
    *   The code now checks `inp.added_sigs` first.
    *   If empty, it falls back to checking `inp.part_sigs`.
    *   If found in `part_sigs`, it retrieves the actual signature bytes using `self.get(sig_proxy)`.
3.  **Signature Retrieval (Multi-Sig):** No changes were needed here, as the existing `self.multisig_signatures(inp)` helper function already correctly aggregates signatures from both `part_sigs` and `added_sigs`.
4.  **Witness Detection:** Slightly refined the `needs_witness` check to consider any SegWit input that has *any* signature (partial or added), ensuring the witness marker/flag is included correctly even if the final signature step didn't involve a SegWit input.
5.  **TXID Calculation:** Ensured the TXID calculation correctly handles the finalized transaction bytes, recalculating body offsets in memory for the SegWit case to avoid potential issues with file descriptor state.

Security Considerations:
*   **Equivalence:** This change maintains the security level of the original code. The core signing logic (`sign_it`) remains unchanged. This fix only addresses the *assembly* phase, ensuring that *all valid, previously generated and stored signatures* are correctly collected and placed into the final transaction structure.
*   **Completeness:** The completeness check is now more robust, correctly verifying that the required number of signatures (1 for single-sig, M for multi-sig) are present, considering all available signature sources within the PSBT input structure. It prevents finalization if any genuinely required signature is missing.
*   **Signature Validity:** The validity of signatures in `part_sigs` is assumed, as they would have been generated by a valid `sign_it` call in a previous step. The `finalize` method's responsibility is aggregation, not re-validation of individual signatures.
*   **No New Vulnerabilities:** This change corrects a state management bug during transaction assembly and does not introduce new cryptographic vulnerabilities or attack surfaces. The process still relies on signatures generated securely by the `sign_it` method.

---

### DIFF
--- a/shared/psbt.py
+++ b/shared/psbt.py
@@ -2258,12 +2258,14 @@ class psbtObject(psbtProxy):
         # - but in segwit case, needs to re-read to calculate it
         # - fd must be read/write and seekable to support txid calc
 
+        start_pos = fd.tell()
         fd.write(pack('<i', self.txn_version))           # nVersion
 
-        # does this txn require witness data to be included?
-        # - yes, if the original txn had some
-        # - yes, if we did a segwit signature on any input
-        needs_witness = self.had_witness or any(i.is_segwit for i in self.inputs if i)
+        # Determine if witness data is needed based on original txn or added segwit signatures
+        needs_witness = self.had_witness or any(
+            inp.is_segwit for idx, inp in enumerate(self.inputs)
+            if (len(inp.part_sigs) + len(inp.added_sigs)) > 0 # Check if *any* sig exists for segwit input
+        )
 
         if needs_witness:
             # zero marker, and flags=0x01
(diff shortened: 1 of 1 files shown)


scgbckbone:
we will mention you in release notes as the person discovering this bug - can we use https://github.com/uebber ?

 Thank you!
I'll provide own patch implementation.

scgbckbone:
fix here https://github.com/Coldcard/firmware/pull/515

PR #515, opened 2025-04-14
finalize foreign single sig outputs from PSBT partial signatures

scgbckbone:


uebber:
Awesome, looks good, thank you!

PR #544, opened 2025-08-26
Clarify allowed usage of Seed XOR standard and name

kdmukai:
There is interest to add Seed XOR support to the project I work on. We already have a work-in-progress PR open in our repo to add the feature.

But there was concern that "Seed XOR" might be a protected / trademarked term.

If Seed XOR is indeed meant to be an open standard, the docs should make this clear.

I view this sort of declaration to be necessary for our work to go forward.

The suggested text here is just my best attempt to communicate that Seed XOR is indeed an open standard and that the name itself may be used in other projects' implementation of this standard.

If there is a better or clearer way to articulate this, I'm happy to amend this PR.

### DIFF
--- a/docs/seed-xor.md
+++ b/docs/seed-xor.md
@@ -157,6 +157,12 @@ with the others on a SEEDPLATE.
     - right to A, down to B ... take that number, and go to that column
     - down to C, that is answer: a &oplus; b &oplus; c
 
+## Open Standard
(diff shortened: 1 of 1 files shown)


kdmukai:
@nvk thank you. I think my suggested text covers that compatibility expectation.

I'll look forward to proceeding with our implementation once this PR is merged.

PR #685, opened 2026-07-11
Feature: Add SLIP-19 support for safe remote-signing of WabiSabi coinjoins

kravens:
# SLIP-19 ownership proofs + HSM support for coinjoin remote signing

Adds what a Coldcard needs to sign coinjoins: SLIP-19 ownership proofs over USB, five policy rules
bounding what unattended signing may do, two screen changes so an unattended device reads honestly,
and one safety fix found while testing on hardware.

Rebased onto `new_edge` (6.6.1X) at doc-hex's suggestion, because Wasabi rounds are taproot and only
the edge line signs P2TR. 4 commits, 13 files, +1079/−13. The 5.6.0-based version this replaces is
kept at tag `slip19-5.6.0-final` and branch `archive/slip19-on-5.6.0` on my fork.

Verified in the simulator on this base; the hardware numbers below were measured on a retail Mk4
running the 5.6.0 version, and the proofs are byte-identical across the two bases.

Happy to split this: the proofs stand alone, and the policy rules could follow as a second PR.

## What it adds

**`slp9` USB command** — a serialized SLIP-19 ownership proof for a derivation path. P2WPKH (ECDSA)
and P2TR (BIP-340 key-spend, BIP-86 tweak). The caller states the address format rather than the
device inferring it from the path purpose, which need not match the script actually used.

The taproot half is why this belongs on edge: proving ownership of a P2TR coin is only useful on a
build that can then sign the round's PSBT. The proof uses this branch's own primitives —
`chains.taptweak()` through `script_pubkey()` for the output key, and `TAP_TWEAK_H` with
`ngu.hash.sha256t()` for the tweak, the same shape `psbt.py` signs key-path inputs with.

**Real SLIP-19 ownership identifiers**, derived per SLIP-21 from the seed, pinned against the
published vectors. An xprv-imported secret has no seed, so proofs are refused rather than given a
fabricated id.

**Two ways to authorise a proof.** Under an HSM policy, `slip19_paths` whitelists the paths and the
proof comes straight back — unattended signing. Without a policy the user approves each proof on
screen, like message signing: `slp9` returns nothing, the screen shows path, address and
SHA-256(commitment), and the host collects the proof with `slok`. Refusing returns `refu`.
So the SLIP-19 user-confirmation flag is only ever set when somebody actually confirmed.

**Five HSM rules bounding unattended signing.** `min_pct_self_transfer` bounds the ratio one
transaction moves; nothing bounded the total, the rate, the price per byte, or whether the round was
worth joining. Each is absent-means-off, shown in the on-screen summary, and in `to_json` so the
policy hash covers it.

| rule | bounds |
| --- | --- |
| `max_txn` | transactions one approved policy may sign, counted per session |
| `max_txn_per_period` | how fast those may be spent, using the existing period |
| `max_sats_leaving` | own value leaving in one transaction, absolute |
| `max_fee_per_kvbyte` | own loss per 1000 vbytes of our own contribution |
| `min_inputs` | fewest inputs the transaction may have, counting every participant |

Why not the existing velocity limits: `per_period` and `max_amount` measure non-change outputs,
which in a coinjoin are the *other participants'* outputs, so any value tight enough to matter
refuses honest rounds. These five measure our own inputs and outputs, or the transaction itself.
`max_fee_per_kvbyte` needs only our own values, because in a coinjoin `calculate_fee()` is `None`
and the transaction-wide fee check is skipped entirely.

## Screen honesty

**Signing indicator.** The HSM screen says "Signing ownership proof" while it works; unattended
signing was otherwise indistinguishable from idle.

**The busy line expires.** If a host stops talking mid-upload nothing raises, so `restore_menu()`
never runs and the screen keeps reading "Receiving..." on an idle device. Nothing is broken — a
fresh `upld` at offset 0 resets the transfer — but an indicator that says "working" when it is not
is the one thing an unattended device must not do. Progress updates refresh the line; 30s of no
movement clears it.

## Safety fix

**Master seed blanking.** `_master_seed` handed back a copy of the 64-byte BIP-39 seed, which
nothing else in the firmware extracts. It is blanked as soon as the one HMAC that needs it has run,
along with each seed-derived intermediate, using the same `blank_object()` the rest of `stash.py`
uses.

## Testing

32 new tests across seven files, on the simulator: proof shapes, determinism, commitment binding,
the HSM path gate, the on-screen approval and refusal, and the ownership id pinned to official
vector 1 (mutation-checked against a wrong SLIP-21 label, the wrong half of the node, and a wrong
root label). Each of the five rules is shown on screen and enforced, absent means off, and composes
with the self-transfer floor; `min_inputs` is mutation-checked against counting only our own inputs.

On this base, with a simulator built from it: `test_slip19.py` 16 passed, and the six rule/screen
suites 20 passed, 36 together.

`test_usb.py` (32 failed, 43 passed, 1 skipped) and `test_hsm.py` (58 failed, 121 passed, 9 errors)
give **identical counts on bare `new_edge` without this branch applied**, checked by rebuilding the
simulator from `38c17eb0` and rerunning. Those failures are my local environment, not this change.

The proofs are byte-identical to the 5.6.0 version for the same seed, path and commitment — P2WPKH
and P2TR both — which is what says the switch to `chains.taptweak()` changed nothing observable.

**On hardware (Mk4).** Proofs accepted at input registration; PSBTs signed unattended under the
policy; coinjoins confirmed on regtest and mainnet, including a round signing five of our own inputs
in one PSBT. Every rule exercised from both sides — refusals at 77%/93.3%/97.8%/98.9%
self-transfer, `max_sats_leaving` at 125,415 against a 100,000 cap, `max_txn_per_period` once the
hourly count was reached, `max_fee_per_kvbyte` at 23,570 against 5,000, `min_inputs` refusing a
4-input round against a floor of 21 and signing a 13-input round against a floor of 3.

## Known limit: signing speed

A coinjoin PSBT costs ~2.7 ms per PSBT byte on an Mk4, end to end, over five mainnet rounds. A
~14 KB PSBT signs in 40s and confirmed on mainnet, but a typical ~40 KB round takes 100–117s and
misses a ~90s coordinator signing phase. The device signs correctly every time; it is simply not
asked early enough.

Nothing here addresses that, and it should not: the cost is in `psbt.py`, which re-deserializes the
unsigned transaction on each of five traversals, and rewriting that touches every signing flow you
have. Flagging it because it decides where unattended coinjoin signing is usable today (~20 KB
PSBTs, ~150 inputs), and a single-pass rewrite looks worth roughly 2x if you ever want it.

## Notes for review

- `hsmcmd` must be enabled for the HSM path; a factory-fresh unit ships with it off. The
  user-approved path needs no policy and no `hsmcmd`.
- Mk4 only for HSM: `version.py` clears `supports_hsm` on Q1. The attended approval path has no such
  restriction, and `test_slip19.py` passes on the Q simulator (12 cases): title bar, 24-char address
  groups and the ENTER/CANCEL prompt all come out right. The policy gate lives in `test_slip19_hsm.py`
  so that it, not the whole module, skips on Q.
- `slp9` is deliberately **not** in `HOBBLED_CMDS`: a proof moves no funds. Say the word if you would
  rather a spending policy blocked it too.
- `approve_transaction` returns early for `por322` PSBTs before the rules loop, so the five counters
  never see one. Coinjoin PSBTs are not `por322`, so I left that alone — but BIP-322 proof-of-reserves
  and SLIP-19 ownership proofs are neighbours, and you may want them reconciled.
- `ckcc-protocol` has no packer for `slp9`/`slok` yet, so hosts use raw `send_recv` today.
- `compute_policy_hash` in `testing/test_hsm.py` mirrors the firmware's `to_json` field order, so
  `slip19_paths` and the five rules had to be added there too, in the same order.
- The busy-line timeout is in `hsm_ux.py`, so it only applies under a policy. The same stale screen
  is reachable outside HSM, but that path has a user present. Say the word if you want it general.

### DIFF
--- a/shared/slip19.py
+++ b/shared/slip19.py
@@ -1,0 +1,62 @@
+# SLIP-19 ownership proofs (BIP-322-style), for coinjoin remote-signing (e.g. Wasabi WabiSabi).
+# Produces a proof a coordinator verifier accepts: a signature over
+#   SHA256( proof_body || cs(scriptPubKey) || scriptPubKey || cs(commitment) || commitment )
+# where proof_body = magic(SL\x00\x19) || flags || varint(count) || 32-byte ownership id(s).
+#
+# The wire result is the full serialized ownership proof: proof_body || bip322_sig
+# (bip322_sig = empty scriptSig (varint 0) || witness stack).
+import ngu, stash
+from public_constants import AF_P2WPKH, AF_P2TR
+
+SLIP19_MAGIC = bytes([0x53, 0x4c, 0x00, 0x19])
+FLAG_USER_CONFIRMATION = 0x01
+
+def _cs(n):
+    # Bitcoin compact-size varint (values used here are small).
+    if n < 0xfd:
+        return bytes([n])
+    return bytes([0xfd, n & 0xff, (n >> 8) & 0xff])
+
+def make_ownership_proof(subpath, flags, commitment):
+    # subpath: str like "m/84h/0h/0h/1/0"; flags: int; commitment: bytes.
+    with stash.SensitiveValues() as sv:
+        node = sv.derive_path(subpath)
+        pk = node.privkey()
+        pubkey = node.pubkey()          # 33-byte compressed
+
+    # scriptPubKey: default Wasabi paths -> P2WPKH for 84h, P2TR for 86h. Decide by path purpose.
+    purpose = subpath.split("/")[1] if "/" in subpath else ""
+    is_taproot = purpose.startswith("86")
+
+    oid = bytes(32)                     # ownership id (see _ownership_id note)
+    proof_body = SLIP19_MAGIC + bytes([flags & 0xff]) + _cs(1) + oid
+
+    if not is_taproot:
+        h160 = ngu.hash.ripemd160(ngu.hash.sha256s(pubkey))
+        spk = bytes([0x00, 0x14]) + h160
+        preimage = proof_body + _cs(len(spk)) + spk + _cs(len(commitment)) + commitment
+        digest = ngu.hash.sha256s(preimage)
+        sig65 = ngu.secp256k1.sign(pk, digest, 0).to_bytes()
+        r = sig65[1:33]
+        s = sig65[33:65]
+        der = _der_sig(r, s) + bytes([0x01])            # + SIGHASH_ALL
+        witness = _cs(2) + _cs(len(der)) + der + _cs(len(pubkey)) + pubkey
+    else:
+        # P2TR: BIP86 output key + BIP-340 schnorr keyspend over the same digest.
+        raise ValueError("taproot slip19 not yet implemented")   # productionization TODO
+
+    bip322_sig = _cs(0) + witness       # empty scriptSig, then witness stack
+    return proof_body + bip322_sig
+
+def _der_sig(r, s):
+    def der_int(x):
+        x = bytes(x)
+        i = 0
+        while i < len(x) - 1 and x[i] == 0:
+            i += 1
+        x = x[i:]
+        if x[0] & 0x80:
+            x = bytes([0]) + x
+        return bytes([0x02, len(x)]) + x
+    body = der_int(r) + der_int(s)
+    return bytes([0x30, len(body)]) + body
--- a/shared/slip19.py
+++ b/shared/slip19.py
@@ -22,21 +22,22 @@ def _tagged_hash(tag, msg):
     th = ngu.hash.sha256s(tag)
     return ngu.hash.sha256s(th + th + msg)
 
-def make_ownership_proof(subpath, flags, commitment):
-    # subpath: str like "m/84h/0h/0h/1/0"; flags: int; commitment: bytes.
+def make_ownership_proof(subpath, addr_fmt, flags, commitment):
+    # subpath: str like "m/84h/0h/0h/1/0"; addr_fmt: AF_P2WPKH or AF_P2TR; commitment: bytes.
+    # The caller states the address format, the same way smsg does. Deriving it from the path
+    # purpose instead would be a guess: the purpose does not have to match the script actually
+    # used at that path, and a proof over the wrong scriptPubKey is silently useless.
+    assert addr_fmt in (AF_P2WPKH, AF_P2TR), 'unsupported address format for ownership proof'
+
     with stash.SensitiveValues() as sv:
         node = sv.derive_path(subpath)
         pk = node.privkey()
         pubkey = node.pubkey()          # 33-byte compressed
 
-    # scriptPubKey: default Wasabi paths -> P2WPKH for 84h, P2TR for 86h. Decide by path purpose.
-    purpose = subpath.split("/")[1] if "/" in subpath else ""
-    is_taproot = purpose.startswith("86")
-
     oid = bytes(32)                     # ownership id (see _ownership_id note)
     proof_body = SLIP19_MAGIC + bytes([flags & 0xff]) + _cs(1) + oid
 
-    if not is_taproot:
+    if addr_fmt == AF_P2WPKH:
         h160 = ngu.hash.ripemd160(ngu.hash.sha256s(pubkey))
         spk = bytes([0x00, 0x14]) + h160
         preimage = proof_body + _cs(len(spk)) + spk + _cs(len(commitment)) + commitment
--- a/shared/slip19.py
+++ b/shared/slip19.py
@@ -17,6 +17,11 @@ def _cs(n):
         return bytes([n])
     return bytes([0xfd, n & 0xff, (n >> 8) & 0xff])
 
+def _tagged_hash(tag, msg):
+    # BIP-340 tagged hash: SHA256( SHA256(tag) || SHA256(tag) || msg ).
+    th = ngu.hash.sha256s(tag)
+    return ngu.hash.sha256s(th + th + msg)
+
 def make_ownership_proof(subpath, flags, commitment):
     # subpath: str like "m/84h/0h/0h/1/0"; flags: int; commitment: bytes.
     with stash.SensitiveValues() as sv:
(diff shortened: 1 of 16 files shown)


kravens:
First mainnet coinjoin with a ColdCard remote-signed input: https://mempool.space/tx/cacadb13ce52cc5b63b24fbb4d95f8a9e2a5d3bb799cf0abfc9e8bc42d5849fc
Doing further tests with a nodebug firmware to confirm everything works safely.

kravens:
Updated / synced on the new base, ready for review (I know you won't have time due to more pressing firmware issues, let me know where I can help). At least ColdCards can still work as USB-connected remote signers with this PR.

shared/slip19.py
```diff
@@ -0,0 +1,139 @@
...
+from public_constants import AF_P2WPKH, AF_P2TR
+
+SLIP19_MAGIC = bytes([0x53, 0x4c, 0x00, 0x19])
+FLAG_USER_CONFIRMATION = 0x01
+
+def _cs(n):
+    # Bitcoin compact-size varint (values used here are small).
+    if n < 0xfd:
+        return bytes([n])
+    return bytes([0xfd, n & 0xff, (n >> 8) & 0xff])
```
doc-hex: please use `serializations:ser_compact_size`

shared/slip19.py
```diff
@@ -0,0 +1,139 @@
...
+# --- SLIP-19 ownership identifier -------------------------------------------------
+# id = HMAC-SHA256(key = k, msg = scriptPubKey), where
+#   k          = Key(m/"SLIP-0019"/"Ownership identification key")   [SLIP-19]
+# derived per SLIP-21:
+#   m          = HMAC-SHA512(key=b"Symmetric key seed", msg=seed)
+#   Child(N,l) = HMAC-SHA512(key=N[0:32], msg=b"\x00" + l)
+#   Key(N)     = N[32:64]
+# The key is cached against the master fingerprint. A different seed - including the same
+# words under a different BIP-39 passphrase - yields a different xfp, so a cached key can
+# never leak across wallets.
```
doc-hex: XFP are usually unique but not always... a collision will make hard to debug failure.

shared/slip19.py
```diff
@@ -0,0 +1,139 @@
...
+_oid_key_cache = None
+
+
+def _master_seed(sv):
```
doc-hex: I would prefer this function to be inlined.

shared/usb.py
```diff
@@ -358,10 +362,18 @@ async def handle(self, cmd, args):
...
         if is_devmode and cmd[0].isupper():
             # special hacky commands to support testing w/ the simulator
+            #
+            # EVAL/EXEC are arbitrary code execution and XKEY injects keypresses, so they must
+            # not also be a way around HSM mode. HSM exists so the device can be left unattended
+            # with a host that may be compromised - which is exactly what unattended coinjoin
+            # signing is. The simulator keeps them, because the test suite drives HSM over this
+            # same path.
+            if hsm_active and not is_simulator():
+                raise HSMDenied
```
doc-hex: remove this code; it does nothing... do_usb_command code is not built due to code in manifest.py

shared/usb.py
```diff
@@ -459,6 +471,41 @@ async def handle(self, cmd, args):
             sign_msg(msg, subpath, addr_fmt)
             return None
 
+        if cmd == 'slp9':
+            # SLIP-19 ownership proof, for coinjoin remote signing (Wasabi WabiSabi).
+            addr_fmt, flags, len_subpath, len_commit = unpack_from('<IIII', args)
+            assert len(args) == (16 + len_subpath + len_commit), 'badlen'
+            from utils import cleanup_deriv_path
```
doc-hex: remainder of this block should move to slip19.py

doc-hex:
Maybe the USB command for the slip-19 proof should be reachable even if HSM not active? It does not require policy settings (right?). If so, it might need user on-screen approval? I don't know if that would be a useful feature by itself.

shared/slip19.py
kravens: Done, also `ser_string` `ser_sig_der` `ngu.hash.hash160` used now. Byte identical results as the handwritten ones, so we should be good.

shared/slip19.py
kravens: Agreed, and this one was worse than a debug problem: on a collision the second wallet would have been handed the first wallet's ownership-identifier key, so its proofs would carry identifiers that do not belong to it.

The cache is now keyed on SHA256(root chain code) instead of the fingerprint. That is 256 bits of the same HMAC output the fingerprint is a truncation of, it is available inside the SensitiveValues block that has to be open anyway, and it changes with the BIP-39 passphrase, so a passphrase wallet cannot collide with its own base wallet either.

Keeping a cache at all is deliberate: every SensitiveValues open re-runs BIP-39 stretching, and a coinjoin round asks for a proof per input.

shared/slip19.py
kravens: Inlined. The three branches (master, words, and the xprv refusal) now sit directly in _ownership_id_key, which is the only thing that ever needed the seed. The blanking try/finally around the one HMAC that consumes it is unchanged.

shared/usb.py
kravens: You are right, and thank you for catching it. usb_test_commands.py is only in the simulator block of manifest.py, so on a real device the import inside that try already fails and the command falls through. The guard could never fire on hardware, which is the only place it claimed to protect.

Removed, along with testing/test_devmode_hsm.py, which only proved the dead branch. usb.py now matches upstream there character for character.

shared/usb.py
kravens: Moved. The handler in usb.py is now four lines: unpack <IIII, check the length, and call slip19.usb_ownership_proof(). Path canonicalisation, the policy gate, the busy-line handling and the new approval flow all live in slip19.py.

The three functions there now take (subpath, addr_fmt, flags, commitment) in that one order. They did not at first, and with four positional arguments, two of them ints and two of them bytes, a transposition would have signed the wrong key without raising.

kravens:
Yeah, I'd probably use it for debugging an attended coinjoin round to follow every step. I'd also be interested to test this user-approved flow on a ColdCard Q, which sadly has no HSM-mode yet (I bricked mine whilst trying to enable it with custom firmware...😭 lesson learned). Here some more notes from Claude about the last commits addressing your points:
> It already was reachable, and you are right that it needs no policy: `slip19_paths` only gates it
> while a policy is active. What it lacked was the on-screen approval, so I have added it.
>
> Outside HSM mode `slp9` now starts a `UserAuthorizedAction` exactly like message signing. The
> device replies with nothing, shows a story naming the derivation path, the address and the
> SHA-256 of the commitment, and the host collects the finished proof with a new `slok` command
> (same poll shape as `smok`). The screen ends with the same "Press OK to continue, otherwise X to
> cancel" line every other approval has, and the text is a module constant like `MSG_SIG_TEMPLATE`.
> Refusing returns `refu` and no proof.
>
> That also fixes something the previous version had to fudge. The SLIP-19 user-confirmation flag
> asserts to the coordinator that a human approved this input; with nobody watching, the old code
> refused to set it. Now a human really does approve, so the flag is honest and is no longer
> refused outside HSM.
>
> Under a policy nothing changes: the approved `slip19_paths` list is the standing consent, and the
> proof still comes straight back without touching the UX, which is what unattended coinjoin
> signing needs.
>
> Useful by itself? Yes, for the attended case: a desktop wallet can register a coin in a round
> with the user confirming each proof on the Coldcard, no HSM policy required. It is slower per
> input, but it is the natural way to try coinjoin signing before committing to HSM mode.
>
> One more host-visible detail: polling `smok` while a proof is pending is refused with
> `Wrong completion command`, and vice versa, so the two completion polls cannot consume each
> other's result.
>
> Note for hosts: the response shape now depends on the mode. Under HSM `slp9` returns the proof;
> otherwise it returns nothing and the caller polls `slok`. `ckcc-protocol` has no packer for
> either command yet, so this is only raw `send_recv` today.

doc-hex:
Thanks for those revisions!

My understanding is Wasabi relies on Tap Root these days. If so, perhaps this change belongs in the "edge" version which can sign TR transactions. To do that, your PR should be rebased against branch `new_edge`.

kravens:
Oh that's true, Wasabi has support for both bech32/native-segwit and taproot. So I agree we should rebase against the Edge firmware.
Also tested the flow of manual ownership proof message signing in the simulator: https://claude.ai/code/artifact/641baa75-fa3c-4c00-9151-837daff4fa10

kravens:
Rebased onto `new_edge`. You are right that this belongs there: proving ownership of a taproot coin is only worth doing on a build that can then sign the round, and Wasabi rounds are taproot.

What changed in the move, beyond the base:

- The taproot path now uses this branch's own primitives instead of my hand-rolled ones — `chains.taptweak()` via `script_pubkey()` for the output key, and `TAP_TWEAK_H` with `ngu.hash.sha256t()` for the tweak, written the way `psbt.py` signs key-path inputs. My `_tagged_hash` helper is gone.
- The rules follow the renames on this base: `sp_idxs` rather than `num_our_keys`, `i.af` rather than `i.addr_fmt`, `msas` rather than `p2sh`, and `slip19_paths` passes `['any']` to `pop_deriv_list` now that `'any'` is no longer built in.
- The 22 commits collapsed to 4 logical ones, since replaying the old history across ~820 commits would have produced a series that neither builds nor reads.

Checked rather than assumed: the proofs are byte-identical to the 5.6.0 version for the same seed, path and commitment, P2WPKH and P2TR both. `test_usb.py` and `test_hsm.py` produce identical pass/fail counts on bare `new_edge` without this branch, so the failures there are my local environment.

Two things I decided one way and would rather you confirm: `slp9` is not in `HOBBLED_CMDS`, on the grounds that a proof moves no funds; and `approve_transaction`'s early return for `por322` means the five counters never see a proof-of-reserves PSBT, which I left alone since coinjoin PSBTs are not `por322`.

The reviewed 5.6.0 version is preserved at tag `slip19-5.6.0-final` and branch `archive/slip19-on-5.6.0` on my fork, so nothing from the earlier review is lost.

PR #693, opened 2026-08-03
bugfix: detect RNG_SR_SEIS and RNG_SR_SECS, retry safely, and fail cl…

scgbckbone:
Compared with #692, this branch is a safer. It follows the documented STM32L4 recovery sequence, validates RNG status both before and after reading data, rejects suspect zero samples, uses bounded retries, catches only the expected keypad OSError, and applies the same protection to the shared Mk4/Q1 bootloader. 

Mk3 support is intentionally deferred to a separate PR.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -4,7 +4,7 @@ This lists the new changes that have not yet been published in a normal release.
 
 # Shared Improvements - Both Mk and Q
 
-- tbd
+- Bugfix: Detect RNG_SR_SEIS and RNG_SR_SECS, retry safely, and fail closed on persistent faults.
 
 # Mk Specific Changes
 
--- a/shared/keyboard.py
+++ b/shared/keyboard.py
@@ -97,7 +97,12 @@ class FullKeyboard(NumpadBase):
     def _start_scan(self):
         # reset and re-start scanning keys
         self.lp_time = utime.ticks_ms()
-        shuffle(self.scan_order)
+        try:
+            shuffle(self.scan_order)
+        except OSError:
+            # An RNG fault may reduce scan-order randomization, but must not
+            # prevent the keyboard from accepting input before login.
+            pass
 
         self._scan_count = 0
         self.waiting_for_any = False
--- a/shared/mempad.py
+++ b/shared/mempad.py
@@ -81,7 +81,12 @@ class MembraneNumpad(NumpadBase):
         # reset and re-start scanning keys
         self.waiting_for_any = False
         self.lp_time = utime.ticks_ms()
-        shuffle(self.scan_order)
+        try:
+            shuffle(self.scan_order)
+        except OSError:
+            # An RNG fault may reduce scan-order randomization, but must not
+            # leave the keypad disabled before the user can log in.
+            pass
 
         self._scan_count = 0
         self._history = bytearray(NUM_ROWS * NUM_COLS)
--- a/stm32/COLDCARD_MK4/rng.c
+++ b/stm32/COLDCARD_MK4/rng.c
@@ -31,6 +31,7 @@
  */
 
 #include <string.h>
+#include <stdbool.h>
 
 #include "py/obj.h"
 #include "py/runtime.h"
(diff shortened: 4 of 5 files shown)


liarco:
**Disclaimer:** I'm a dev by trade, but I don't have years of experience in low-level or programming or in developing software for this kind of devices. So please bare with me if I'm making mistakes in my message below. I'm a CCQ owner and have skin in the game with the ongoing issues, so I'm doing my best to follow along with the changes to this repo and I would like to (a) understand what's being added to this codebase better (b) contribute to the review process in the hope that it can be useful.

Hi @scgbckbone,
while reviewing this PR I noticed that, unlike #692, this implementation checks the seed error flags (`RNG_SR_SEIS` and `RNG_SR_SECS`) but not the clock error flags (`RNG_SR_CEIS` and `RNG_SR_CECS`).

From ST's HAL documentation I understand that a clock error does not invalidate values generated before the error occurred, so I can see why seed errors and clock errors are treated differently. However, I'm not sure whether the documentation guarantees that a value read while CEIS/CECS are set is always safe to consume, or whether these flags should also influence the recovery path.

More generally, if the hardware reports a clock error, should the firmware explicitly acknowledge and recover from that condition (or at least surface it somehow), rather than ignoring it?

I'm asking because my intuition is that if the hardware exposes dedicated clock-error flags, they're probably worth handling explicitly, but I may well be missing some STM32-specific behaviour here.

Thank you for your time.

liarco:
One additional thought.

I completely understand that this PR is probably intended as a quick mitigation, especially because some users may currently be locked out of their devices. In that context, preserving usability makes a lot of sense.

That said, I'm a little concerned about silently falling back to a deterministic scan order if the TRNG repeatedly fails. The keypad scan randomization exists as a security mitigation, so if it is intentionally being disabled, I wonder whether the user should be informed somehow.

Personally, I would feel more comfortable if the device displayed a warning (or otherwise recorded the condition) so I know that one of the security features is currently unavailable. That would let me decide whether I want to continue using the device or stop and investigate the hardware issue.

The Coldcard can still perform many useful operations even if the TRNG is temporarily unavailable, so I don't think it necessarily needs to become unusable. I just think users should know when one of its security mitigations is no longer active.

scgbckbone:
> **Disclaimer:** I'm a dev by trade, but I don't have years of experience in low-level or programming or in developing software for this kind of devices. So please bare with me if I'm making mistakes in my message below. I'm a CCQ owner and have skin in the game with the ongoing issues, so I'm doing my best to follow along with the changes to this repo and I would like to (a) understand what's being added to this codebase better (b) contribute to the review process in the hope that it can be useful.
> 
> Hi @scgbckbone, while reviewing this PR I noticed that, unlike #692, this implementation checks the seed error flags (`RNG_SR_SEIS` and `RNG_SR_SECS`) but not the clock error flags (`RNG_SR_CEIS` and `RNG_SR_CECS`).
> 
> From ST's HAL documentation I understand that a clock error does not invalidate values generated before the error occurred, so I can see why seed errors and clock errors are treated differently. However, I'm not sure whether the documentation guarantees that a value read while CEIS/CECS are set is always safe to consume, or whether these flags should also influence the recovery path.
> 
> More generally, if the hardware reports a clock error, should the firmware explicitly acknowledge and recover from that condition (or at least surface it somehow), rather than ignoring it?
> 
> I'm asking because my intuition is that if the hardware exposes dedicated clock-error flags, they're probably worth handling explicitly, but I may well be missing some STM32-specific behaviour here.
> 
> Thank you for your time.

Thanks for reviewing! this is intentional: RM0432 states that a clock error does not invalidate generated random numbers, so a word with DRDY set remains usable. Unlike a seed error, it does not require RNG recovery. If no word becomes available, the existing timeout still fails. CEIS is left set because RNG interrupts are disabled- clearing it is irrelevant. I added comments  c608e96d

scgbckbone:
> I completely understand that this PR is probably intended as a quick mitigation, especially because some users may currently be locked out of their devices

we found no evidence that already shipped RNG patch can brick user devices. https://x.com/AVirgovic/status/2084307173480276260?s=20 (so no need to hurry with this)

scgbckbone:
Pushed fb5e69d4 — `rng_recover()` now implements the full RM0432 §32.3.7 seed-error recovery: after clearing SEIS and cycling RNGEN, it reads and discards 12 words from RNG_DR to flush pre-error residue from the pipeline.                                           
                                                                                                                                                             
The flush bails early on a recurring seed error (or DRDY timeout in the main firmware); retry policy stays with the existing attempt loop, and recovery never consumes a word, so nothing fails open. RM step 3 (confirm SEIS clear) is already covered by the mask checks in r`ng_try_once()` / `rng_sample()`.

liarco:
> Thanks for reviewing! this is intentional: RM0432 states that a clock error does not invalidate generated random numbers, so a word with DRDY set remains usable. Unlike a seed error, it does not require RNG recovery. If no word becomes available, the existing timeout still fails. CEIS is left set because RNG interrupts are disabled- clearing it is irrelevant. I added comments [c608e96](https://github.com/Coldcard/firmware/commit/c608e96d5d8be23b949e76fe2fcd0b14047791c0)

Thank you for taking the time to reply, @scgbckbone.
The explanation and the comments in the code are clear and much appreciated.

> we found no evidence that already shipped RNG patch can brick user devices. https://x.com/AVirgovic/status/2084307173480276260?s=20 (so no need to hurry with this)

Didn't mean to overstate the effect, thanks for clarifying. In normal conditions (no faulty HW) the device should recover after a power-cycle.

PR #694, opened 2026-08-03
rng: boot-time self-test proving rng_get() enters the hardware read path

scgbckbone:


scgbckbone (commented):
requires https://github.com/Coldcard/firmware/pull/693

scgbckbone (commented):
OK by KIMI3

doc-hex:
I like the test, but I would remove the code to print the "pass" msg; one, because it cannot never not be printed (and crash is appropriate response to broken RNG), and two, it cannot be seen as the serial port output is extremely limited in production devices. ... the need to print the fixed text introduces so much complexity that can be removed.

PR #696, opened 2026-08-03
Mk3: RNG self-test proving rng_get() enter the hardware read path

scgbckbone:
* same as https://github.com/Coldcard/firmware/pull/694 just for Mk3
* requires https://github.com/Coldcard/firmware/pull/695


scgbckbone (commented):
OK by KIMI3

PR #698, opened 2026-08-04
rng: symlink COLDCARD/rng.c to the Mk4 file instead of duplicating it

Silexperience210:
Replaces the original contents of this PR. @geldot is right: after #693 the two files
are identical apart from comments, and they were byte-identical on `master` before either
PR touched them. `COLDCARD_Q1/rng.c` is already a symlink to `../COLDCARD_MK4/rng.c`, so
this just applies the pattern already in the tree.

Two corrections to what I originally filed here, both worth stating plainly:

**This branch does not build the Mk3.** `MK-Makefile` sets `BOARD=COLDCARD_MK4` and
`Q1-Makefile` sets `BOARD=COLDCARD_Q1`; there is no Mk3 target. So `stm32/COLDCARD/rng.c`
on `master` is never compiled, and my original commit here was patching a file that does
not reach any device. The Mk3 code that shipped as 4.2.0 lives on `v4-legacy`. Filed
separately as a PR against that branch.

**My RM0351 retarget was not verified.** The only substantive thing I changed relative to
#693 was the reference. The 12-word pipeline flush is RM0432 wording; the older L4-series
manual documents the shorter clear-SEIS-then-toggle-RNGEN sequence. I retargeted the
citation by search-and-replace without checking that the newer procedure appears there,
which was sloppy. Collapsing to a symlink removes the question entirely.

So: tree hygiene only, no functional change. Close it if you would rather keep the two
files separate for future divergence.


scgbckbone:
already here https://github.com/Coldcard/firmware/pull/695


This PR is opened against wrong branch - close pls and review 695 if possible - thanks

geldot:
Unless I'm missing something, the CC and CC4 `rng.c` source files are now identical (except for the comments) post #693 / #698. If so, there is no point in keeping two copies in tree as though they were distinct.

```diff
--- rng-cc.c    2026-08-04 13:50:08.269177951 +0000
+++ rng-cc4.c   2026-08-04 13:50:35.431394302 +0000
@@ -54,8 +54,8 @@
 #define RNG_TIMEOUT_MS      (10)
 #define RNG_MAX_ATTEMPTS    (3)

-// Clock-error flags (CEIS/CECS) are intentionally ignored: per RM0351
-// "Error management", "the clock error has no impact on generated random
+// Clock-error flags (CEIS/CECS) are intentionally ignored: per RM0432
+// section 32.3.7, "the clock error has no impact on generated random
 // numbers", and ST's errata (ES0250/ES0335) confirm a clock error neither
 // stops generation nor invalidates RNG_DR when DRDY is set. A dead clock
 // still fails closed via the DRDY timeout below. CEIS is left set on
@@ -75,7 +75,7 @@
     RNG->CR &= ~RNG_CR_RNGEN;
     RNG->CR |= RNG_CR_RNGEN;

-    // RM0351 "Error management": after clearing SEIS, read out 12 words from RNG_DR and
+    // RM0432 32.3.7: after clearing SEIS, read out 12 words from RNG_DR and
     // discard each of them to clean the pipeline of pre-error residue.
     // Bounded: if the error recurs or DRDY stops arriving, bail out and let
     // the next attempt's flag checks and DRDY timeout handle it.
```

Silexperience210:
Good catch, thank you — rewritten as a symlink.

Chasing it down turned up the bigger problem: `master` has no Mk3 build target, so the
file I was originally patching here is never compiled. The Mk3 fix belongs on `v4-legacy`,
which is where 4.2.0 was built from. Opened there instead.

Also correcting myself on the reference: my RM0351 retarget was a search-and-replace, and
the 12-word flush is RM0432 wording that I did not verify against the older manual. The
symlink makes that moot.

Silexperience210:
Apologies — I missed your comment above before pushing the rewrite, and answered @geldot's
without reading yours first. #695 already covers the Mk3 on the right branch, including the
bootloader and the changelog, so closing this as you asked.

Closing #699 too. I opened it against `v4-legacy` this afternoon without checking for
existing PRs, so it duplicates #695 and covers less. Entirely my error, and it cost you a
review cycle.

scgbckbone:
Np, thanks for contributing.

PR #699, opened 2026-08-04
rng: seed-error recovery for Mk3 (port of #693 to v4-legacy)

Silexperience210:
#693 fixes Mk4/Mk5/Q. This is the same change for the Mk3, filed against the branch that
actually builds it — `master` has no Mk3 target, so the copy of `stm32/COLDCARD/rng.c`
there is never compiled, while this branch carries the code that shipped as 4.2.0 on
2026-07-31.

The file here is byte-identical to master's and has the same gap: `rng_init()` only tests
`RNGEN`, but a seed error latches `SEIS` and stops `DRDY` while leaving `RNGEN` set, so the
peripheral is never recovered and every subsequent `rng_get()` times out for the rest of
the boot session. Since the hotfix wired `rng_get()` to the hardware TRNG, that path now
reaches the keypad scan-order shuffle in `_start_scan()`, which runs from a `Pin.irq`
callback before login.

`shared/mempad.py` gets the same `try/except OSError` as #693. This branch has no
`keyboard.py`, so there is nothing else to guard.

**On the reference:** the Mk3 is an STM32L475, covered by RM0351, which documents the
shorter clear-SEIS-then-toggle-RNGEN sequence. The 12-word pipeline flush is RM0432
wording. I kept it, with a note in the header, so this file stays identical to the Mk4/Q
version rather than quietly diverging — the extra discards are harmless. Say the word if
you would rather I trim it to the RM0351 sequence.

Should go in after #693. I have no Mk3 to test on, so this needs a look from someone who
does.

### DIFF
--- a/shared/mempad.py
+++ b/shared/mempad.py
@@ -77,7 +77,13 @@ class MembraneNumpad(NumpadBase):
         # reset and re-start scanning keys
         self.waiting_for_any = False
         self.lp_time = utime.ticks_ms()
-        shuffle(self.scan_order)
+        # Scan-order randomisation is anti-Tempest hygiene, not a secret.
+        # rng_get() can now raise, and this runs from an IRQ callback before
+        # login, so a TRNG hiccup must not be able to take the keypad down.
+        try:
+            shuffle(self.scan_order)
+        except OSError:
+            pass
 
         self._scan_count = 0
         self._history = bytearray(NUM_ROWS * NUM_COLS)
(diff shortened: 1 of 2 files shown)


Silexperience210:
Closing — duplicate of #695, which predates this and is more complete (it also covers
`stm32/bootloader/rng.c` and the changelog). I opened this without checking existing PRs
first. Sorry for the noise.

The only thing here that isn't in #695 is a standalone testbench that mocks RNG_CR/SR/DR
and exercises the driver against nominal / transient seed error / dead 48 MHz clock /
1-in-5 glitch, if it's of any use for regression:
https://github.com/Silexperience210/firmware/tree/rng-fault-analysis — no obligation.

PR #700, opened 2026-08-05
rng: verify libngu bytes path reaches hardware driver

scgbckbone:


scgbckbone (commented):
KIMI3:
- rng-code-check now requires relocations to `checked_chip_trng_read` (and a `.text.checked_chip_trng_read` section) in libngu/random.o, but the libngu submodule pinned by this branch (537519a8) has no such symbol — `my_random_bytes()` there calls `rng_get()` directly via the `CHIP_TRNG_32()` macro. `make all` will fail at this check on every board until the PR also bumps external/libngu to a commit that contains `checked_chip_trng_read`.

stm32/shared.mk
```diff
@@ -74,6 +74,31 @@ rng-code-check:
...
+	if ! printf '%s\n' "$$random_bytes_refs" \
+		| grep -Eq '[[:space:]]R_ARM_(THM_)?CALL[[:space:]]+my_random_bytes([+-].*)?$$'; then \
+		echo "ERROR: ngu.random.bytes does not call my_random_bytes"; \
+		printf '%s\n' "$$random_bytes_refs"; \
+		exit 1; \
+	fi; \
+	my_random_bytes_refs="$$($(OBJDUMP) -r -j .text.my_random_bytes \
+		"$$libngu_random")" || exit $$?; \
+	if ! printf '%s\n' "$$my_random_bytes_refs" \
+		| grep -Eq '[[:space:]]R_ARM_(THM_)?CALL[[:space:]]+checked_chip_trng_read([+-].*)?$$'; then \
```
scgbckbone, line 90: This check cannot pass against the libngu submodule pinned on this branch (external/libngu @ 537519a8): that revision's `ngu/random.c` defines no `checked_chip_trng_read` at all — `my_random_bytes()` calls `rng_get()` directly via the `CHIP_TRNG_32()` macro, so `.text.my_random_bytes` contains no `R_ARM_*CALL checked_chip_trng_read` relocation and the build fails here. Same issue at line 95: the `.text.checked_chip_trng_read` section will not exist in `libngu/random.o`. The check itself looks correct for the newer libngu layout (e.g. switck/libngu@5d04106a, where `random_bytes -> my_random_bytes -> checked_chip_trng_read -> chip_trng_read -> rng_get` does hold), but this PR must also bump the submodule (or land after the bump), otherwise merging breaks `make all` for MK4/Q1.

scgbckbone: yes - this requires https://github.com/switck/libngu/commit/5d04106a

scgbckbone:
- [x] requires https://github.com/switck/libngu/commit/5d04106a libngu - so https://github.com/Coldcard/firmware/pull/713 is expected to be merged first

PR #701, opened 2026-08-05
bugfix: harden callgate buffer bounds

scgbckbone:
## Summary

- validate the complete caller-provided buffer span in the Mk4/Q bootloader callgate
- use subtraction-based bounds checks so address-plus-length arithmetic cannot wrap
- reject negative callgate lengths explicitly
- require the full 12-byte output buffer used by the MCU-key usage method
- document the bugfix in releases/Next-ChangeLog.md

## Security impact

The Mk4/Q SRAM check previously calculated the end address as x + len, which could wrap. Its read-only firmware check validated only the starting address.

The replacement checks first establish a non-negative offset inside the permitted region and then compare the requested length with the remaining region size. No unchecked end-address addition is performed.

This hardening follows the callgate bounds observation discussed in the Karma-X review: https://karma-x.io/blog/post/75/

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -15,6 +15,8 @@ This lists the new changes that have not yet been published in a normal release.
   Thanks to [@instagibbs](https://github.com/instagibbs) for reporting this issue.
 - Bugfix: Fixed PSBT uploads being mistaken for partial firmware uploads.
 - Bugfix: Prevent valid message signatures when using a Delta Mode PIN.
+- Bugfix: Harden callgate buffer validation against integer overflow and out-of-range access,
+  following a finding in the [Karma-X security review](https://karma-x.io/blog/post/75/).
 
 # Mk Specific Changes
 
--- a/stm32/mk4-bootloader/dispatch.c
+++ b/stm32/mk4-bootloader/dispatch.c
@@ -33,6 +33,19 @@
 #include "stm32l4xx_hal.h"
 
 
+// range_is_inside()
+//
+    static bool
+range_is_inside(uint32_t addr, uint32_t len, uint32_t base, uint32_t size)
+{
+    if(addr < base) return false;
+
+    uint32_t offset = addr - base;
+
+    // Subtraction-based bounds checks avoid overflow in addr + len.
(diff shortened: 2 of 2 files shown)


scgbckbone (commented):
OK by KIMI3

doc-hex:
(from Codex)
No blocking correctness findings. Verdict: technically sound; approve after its changelog conflict is rebased.

Key checks:

- Cortex-M4 ABI is compatible: pointers, `int`, `size_t`, and `uint32_t` are all 32-bit. `3 * sizeof(uint32_t)` correctly covers the three `int` stores.
- The caller passes length in `r2`; the callee interprets the same 32 bits as signed `int`. GCC correctly compiles the `0 <= len <= 1024` check as one unsigned comparison, rejecting negative values too.
- The subtraction-based range math is correct and cannot underflow or overflow with the current regions:
  - SRAM: `[0x20000000, 0x2009e000)`
  - Firmware: `[0x08020000, 0x08200000)`
- Typical callgate behavior is unchanged:
  - Current bytearray buffers are positive-sized, SRAM-resident, and at most 696 bytes.
  - `None` calls still pass `(NULL, 0)` and never invoke `good_addr()`.
  - Method 25 already supplies 12 bytes; increasing its minimum from 8 to 12 only closes the existing out-of-bounds case.
- One intentional edge change: a zero-length range located exactly at a region’s exclusive end is now rejected. No current `REQUIRE_*` call has a zero-byte minimum, so this does not affect existing callers.
- `uint32_t` instead of `uintptr_t`, and `sizeof(uint32_t)` alongside `int *`, are target-specific rather than portable—but valid for this Cortex-M4/ARM EABI build.

Validation performed:

- Compiled the changed translation unit with the actual Cortex-M4 flags for both Mk4 and Q1 without warnings.
- Inspected generated Thumb assembly for the signed-length and range checks.
- Full bootloader linking was blocked by the unrelated missing Python `namedlist` build dependency.
- There are no automated boundary tests in the PR; adding cases for wrapped SRAM addresses, firmware-end crossing, negative length, and method 25’s 8/12-byte boundary would be worthwhile but not blocking.

PR #702, opened 2026-08-05
bugfix: invalidate message signatures in Delta Mode

scgbckbone:
## Summary

- Silently invalidate seed-derived message signatures when logged in with a Delta Mode PIN, matching transaction-signing behavior.
- Keep explicitly supplied private-key/WIF signing behavior unchanged.

## Security impact

Transaction signing already double-hashes the digest in Delta Mode so signatures are invalid for the intended transaction. Message signing did not apply the same protection and produced a valid signature with the real seed-derived key. Applying the guard in `sign_message_digest()` covers the USB, SD, NFC, QR, and export-signing paths that use the central primitive.


scgbckbone (commented):
OK by KIMI3

PR #703, opened 2026-08-05
bugfix: short PSBT uploads at firmware probe boundary

scgbckbone:
## Summary
- only inspect the firmware header after receiving the full probe block
- cover PSBT uploads ending immediately after the probe boundary


scgbckbone (commented):
OK by KIMI3

PR #704, opened 2026-08-05
block SIGHASH_SINGLE by default

scgbckbone:
just block it - if someone want to use it - then can disable sighash checks


scgbckbone (commented):
OK by KIMI3

PR #705, opened 2026-08-05
bugfix: separate SE1 check nonce

scgbckbone:
needs physical access after you insert PIN

### DIFF
--- a/stm32/mk4-bootloader/se2.c
+++ b/stm32/mk4-bootloader/se2.c
@@ -1195,6 +1195,7 @@ se2_encrypt_secret(const uint8_t secret[], int secret_len, int offset,
 
     if(check_value) {
         // encrypt the check value: 32 zeros
+        nonce[15] = 0x80;       // separate counter range from encrypted secrets
         aes_init(&ctx);
         ctx.num_pending = 32;
         aes_done(&ctx, check_value, 32, aes_key, nonce);
@@ -1239,6 +1240,7 @@ se2_decrypt_secret(uint8_t secret[], int secret_len, int offset,
 
     if(check_value) {
         // decrypt the check value
+        nonce[15] = 0x80;       // separate counter range from encrypted secrets
         aes_init(&ctx);
         aes_add(&ctx, check_value, 32);
         uint8_t got[32];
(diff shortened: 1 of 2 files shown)


stm32/mk4-bootloader/se2.c
```diff
@@ -1239,6 +1240,7 @@ se2_decrypt_secret(uint8_t secret[], int secret_len, int offset,
 
     if(check_value) {
         // decrypt the check value
+        nonce[15] = 0x80;       // separate counter range from encrypted secrets
```
scgbckbone, line 1243: Existing devices wrote the check value with the legacy nonce (nonce[15] = offset/16 = 0, since the only writer, pins.c:973, uses offset=0). After a firmware upgrade this decrypt with nonce[15] = 0x80 will fail check_all_zeros() for every wallet created by older firmware, so se2_decrypt_secret() returns is_valid=false and pin_fetch_secret() zeroes the secret and sets PA_ZERO_SECRET (pins.c:1089-1098) — users see a blank wallet even though the encrypted seed in SE1 is intact and would still decrypt fine with counter 0. This needs a migration path (e.g. on check failure, retry with the legacy nonce, and if that validates, rewrite the check value in the new format) or at least an explicit documented recovery story, otherwise every existing Mk4/Q1 user loses access to their secret on upgrade (permanently, if they have no external seed backup).

scgbckbone (commented):
KIMI3:
- No migration/backward-compat path for existing check values: all deployed devices stored KEYNUM_check_secret encrypted with the legacy nonce (counter 0, same range as the secret). Decrypting it with nonce[15]=0x80 makes the check fail, so pin_fetch_secret() reports a blank wallet after upgrade and the still-intact seed becomes unreadable by the new firmware.

stm32/mk4-bootloader/se2.c
doc-hex: Bootloader code is not field upgradable, so I don't think migration would be an issue.

scgbckbone: correct - not an issue

doc-hex:
I would like to look at the datasheet and test on real hardware before merging this change.

doc-hex:
I understand the issue and the fix better now and I'm comfortable merging this.

PR #706, opened 2026-08-05
bootloader: reject DFU images longer than signed firmware_length

scgbckbone:
verify_firmware_in_ram() authenticated only hdr->firmware_length bytes, but the untrusted DFU element length was passed unchecked to psram_do_upgrade(), which flashed that many bytes. Anything appended past the signed length (up to FW_MAX_LENGTH_MK4, which reaches the top of flash and covers the whole internal /flash filesystem) was written to flash without possessing the signing key, e.g. via recovery-mode SD upgrade of the currently-installed signed image.

- verify_firmware_in_ram(): require len == hdr->firmware_length, fail closed on any mismatch
- psram_do_upgrade(): assert size <= FW_MAX_LENGTH_MK4 (defense in depth)

### DIFF
--- a/stm32/mk4-bootloader/psram.c
+++ b/stm32/mk4-bootloader/psram.c
@@ -308,6 +308,7 @@ psram_recover_firmware(void)
 psram_do_upgrade(const uint8_t *start, uint32_t size)
 {
     ASSERT(size >= FW_MIN_LENGTH);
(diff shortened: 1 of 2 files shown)


scgbckbone (commented):
OK by KIMI3

PR #707, opened 2026-08-05
Improve seed gen & require external entropy for each new wallet

scgbckbone:
* each seed creations ask for randomness both SE1/SE2
* no new wallet can be created with TRNG only (external randomness required)
* add coin flips as option
* add key mashing as option
* kept dice as an option too
* remove showing hash of dicerolls from TRNG + dice (only show hash in Advanced dice only)
* reseeding with full 32 bytes
* updated libngu with SHA256 DRBG instead of yasmarang

Thanks to [Roasbeef](https://github.com/Roasbeef) for key mashing idea https://x.com/roasbeef/status/2083265977048059988?s=20


external/libngu
scgbckbone, line 1: one issue here: 

as we updated libngu to master HEAD, it now pulls lot of code that we do not need in master (not EDGE) release - musig/schnorr. I prefer we merge https://github.com/switck/libngu/pull/56 which allows us not bundle it 

@doc-hex

scgbckbone:
latest commit https://github.com/Coldcard/firmware/pull/707/commits/f4c35993f9548f17f8cb6fa22fd9fd830e788c14 requires https://github.com/switck/libngu/pull/68

- [ ] After above is merged - submodule needs to be bumped

scgbckbone:
replaced by strictly better https://github.com/Coldcard/firmware/pull/713

PR #709, opened 2026-08-06
bugfix: WIF Store duplicate detection after restart

scgbckbone:
WIF settings are persisted as JSON arrays. Previously, both import paths created `(pubkey, privkey)` tuples, while a settings reload restored those records as lists. The membership check therefore failed to recognize an already stored WIF after restart.

This creates JSON-native `[pubkey, privkey]` records from the start. The regression performs a device-side `ujson.dumps()`/`ujson.loads()` round-trip before reimporting the same WIFs and confirms that the store remains unchanged.

Also adds the bugfix to `releases/Next-ChangeLog.md`.

Tested:
- `test_wif_store_import_duplicate`


scgbckbone (commented):
OK by KIMI3

PR #715, opened 2026-08-10
bugfix: wipe in deltamode before listing saved passphrases

scgbckbone:
`PassphraseSaverMenu.construct()` calls `_calc_key()`, which reads the real master secret via `SensitiveValues(bypass_tmp=True)` with no `enforce_delta`. In Delta Mode the secret is genuine, so the derived key decrypts the saved-passphrase file and the menu shows saved passphrases by partial text (full text when no prefix disambiguates), each wallet's XFP, and a Restore item.

Seed Vault guards the equivalent data with `wipe_if_deltamode()`; this does the same before the file is read.

Gate is on the menu, not `_calc_key()`, because `MicroSD2FA` inherits `_calc_key()` for the login-time 2FA check.

Tested on simulator: test_pwsave (11 passed). Delta Mode path itself not exercised - needs SE2 trick PIN setup.


scgbckbone (commented):
OK by KIMI3

PR #716, opened 2026-08-10
Fix nested passphrase wallet display and backups

scgbckbone:
## Summary

BIP-39 passphrase wallets are represented internally as temporary extended private keys. Because a passphrase can be applied to a word-based temporary seed, the persistent master seed is not necessarily the passphrase wallet’s parent.

Update secret display and backup workflows to consistently use the wallet currently in effect:

- Show the effective XPRV in View Seed Words when a BIP-39 passphrase is active, without displaying the parent seed words or passphrase.
- Make Backup System, Clone Coldcard, and Key Teleport’s Full COLDCARD Backup capture the active temporary or passphrase wallet.
- Warn before exporting when a temporary seed or passphrase wallet is active.
- Remove the choice between master and passphrase-wallet backups, since that choice cannot correctly represent passphrases applied to temporary seeds.
- Simplify related temporary-secret bypass and settings-switching logic.
- Correct related passphrase restoration and temporary-to-master messaging.

The backup file format is unchanged. Passphrase-wallet backups contain the effective XPRV, not the parent seed words or passphrase.

### DIFF
--- a/docs/backup-files.md
+++ b/docs/backup-files.md
@@ -41,13 +41,14 @@ called `ckcc-backup.txt`, but the filename is now picked randomly.
 ## BIP39 Passphrase
 
 If BIP39 passphrase is active, the passphrase wallet itself is backed-up,
-as extended private key created from seed words plus passphrase. Neither
-the seed words nor the passphrase are part of such backup.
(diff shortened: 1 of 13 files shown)


scgbckbone (commented):
OK by KIMI3

PR #717, opened 2026-08-10
cleanup: drop unused fetch() params

scgbckbone:
`PinAttempt.fetch()` had `duress_pin` and `spare_num` parameters that no caller passes.

- `duress_pin` is the mk3-and-earlier duress path
- `spare_num` branch returning a zeroed buffer while a temporary seed is active was unreachable

`roundtrip(4)` is equivalent to the old `roundtrip(4, spare_num=0)`: `spare_num=None` skips the block, and `0 << 8` was a no-op on `change_flags`.

Tested on simulator: test_pwsave (11 passed), test_bip39pw (88 passed).

### DIFF
--- a/shared/pincodes.py
+++ b/shared/pincodes.py
@@ -361,20 +361,12 @@ class PinAttempt:
         # - call new_main_secret() when main secret changes!
         # - is_secret_blank and is_successful may be wrong now, re-login to get again
 
-    def fetch(self, duress_pin=None, spare_num=0, bypass_tmp=False):
+    def fetch(self, bypass_tmp=False):
         if self.tmp_value and not bypass_tmp:
             # must make a copy here, and must be mutable instance so not reused
-            if spare_num:
-                return bytearray(AE_SECRET_LEN)
             return bytearray(self.tmp_value)
 
-        if duress_pin is None:
-            secret = self.roundtrip(4, spare_num=spare_num)
-        else:
-            # mk3 and earlier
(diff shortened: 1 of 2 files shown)


shared/pincodes.py
```diff
@@ -361,20 +361,12 @@ def change(self, **kws):
         # - call new_main_secret() when main secret changes!
         # - is_secret_blank and is_successful may be wrong now, re-login to get again
 
-    def fetch(self, duress_pin=None, spare_num=0, bypass_tmp=False):
+    def fetch(self, bypass_tmp=False):
```
scgbckbone, line 364: Dropping `duress_pin` here leaves `testing/test_pincodes.py::test_duress` (gated by `only_mk3`) broken: line 128 there still does `pa.fetch(duress_pin=b'34-34')`, which will now raise `TypeError: unexpected keyword argument` whenever the mk3 simulator suite runs. The `spare_num` removal is safe (no caller passes it, and `roundtrip(4)` is equivalent to the old `roundtrip(4, spare_num=0)` since marshal treats `0 << 8` as a no-op), but the stale test should be updated or removed in this PR.

scgbckbone (commented):
KIMI3:
- Removing the `duress_pin` parameter breaks `testing/test_pincodes.py::test_duress` (mk3-only), which still calls `pa.fetch(duress_pin=b'34-34')` — that assertion now raises TypeError; the test should be updated or removed in this PR.

PR #719, opened 2026-08-12
bugfix: dwld served PSRAM that the current USB session never staged

drk1wi:
`handle_download` (`shared/usb.py:757-784`) bounds the read by the size of the window
and never by what this session actually put there:

```python
        assert 0 <= offset < MAX_TXN_LEN, "bad offset"
        assert offset + length <= MAX_TXN_LEN, "bad offset"
        ...
        pos = (MAX_TXN_LEN * file_number) + offset
        PSRAM.read(pos, buf)
```

Between them the two file numbers cover the whole 4MB window (`psram.py:10-11`), and
the lower half is never cleared: `psram_wipe_and_setup` (`psramdisk.c:527`) memsets
`PSRAM_TOP_BASE`, the upper half. So a later connection reads back what an earlier one
left, which for the lower window is the staged PSBT (`usb.py:848`) and the signed
result (`auth.py:779`). A plaintext link that never sends `ncry` swept all 4,194,304
bytes over 2,048 requests, and a separate session with its own `ncry` read another
session's PSBT and signed result with checksums matching the receipts the device had
handed out.

Fix: track how much of each of the two files this session staged and serve only that.
`self.readable = [0, 0]` clears at offset zero of an `upld` and rises after the write,
and is set from the length the device itself reports in the `stok`/`bkok` branch.
`handle_crypto_setup` clears both, and `dwld` now requires the encrypted session so a
peer that never sends `ncry` cannot inherit the previous session's extents.
`USBHandler` is a singleton built once per power-up (`usb.py:80-107`), so `ncry` is the
only point at which per-session state can be reset.

The file number comes from `isinstance(req, RemoteBackup)` rather than the command
name, because `stok` and `bkok` both reach that branch against whatever request is
pending (`:544`), so the host would otherwise choose which extent gets opened. This
matches how `:392-396` already keys on `isinstance(req, FirmwareUpgradeRequest)`.

Not covered: `MAX_UPLOAD_LEN` is 4MB and `upld` writes at absolute `pos` (checked at
`:801`), so an upload running past 2MB lands in file one's window while only file
zero's extent rises. The only upload I found that goes that large is the firmware
upgrade, which is checked with `sha2` rather than read back, so I left it rather than
guess at the right accounting.

---

### DIFF
--- a/shared/usb.py
+++ b/shared/usb.py
@@ -135,6 +135,11 @@ class USBHandler:
         self.file_checksum = sha256()
         self.is_fw_upgrade = False
 
+        # How much of each of the two PSRAM files this session has staged.
+        # - 'dwld' serves nothing past this, so a session cannot read back
+        #   what some other session left in PSRAM
+        self.readable = [0, 0]
+
         # handle simulator
         self.blockable = getattr(self.dev, 'pipe', self.dev)
 
@@ -544,7 +549,7 @@ class USBHandler:
         if cmd == 'stok' or cmd == 'bkok' or cmd == 'smok' or cmd == 'pwok':
             # Have we finished (whatever) the transaction,
             # which needed user approval? If so, provide result.
-            from auth import UserAuthorizedAction
+            from auth import UserAuthorizedAction, RemoteBackup
 
             req = UserAuthorizedAction.active_request
             if not req:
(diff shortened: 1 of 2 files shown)


doc-hex:
looking good. please add a line to `Next-ChangeLog.md`

scgbckbone (commented):
(comment withdrawn by author - posted by mistake)

scgbckbone (commented):
nice find on the plaintext sweep...                                                                                              
                                                                                                                                                             extent accounting has a leak: partial `upld` at nonzero offset (no offset 0 first) grows readable[0] past what the session wrote. upload 256 bytes at offset 256 in a fresh session and `dwld` 0..512 - first half is the previous session's data.                                                                                                         
                                                                                                                                                             
alternative in #724: only device produced results are downloadable, any `upld` block or new staged txn kills the lease. no counters to drift. kept your two good ideas: `dwld` needs encryption adn  `ncry` voids the lease

take a look at #724 if you have a minute, thanks

drk1wi:
@scgbckbone Thank you, and you are absolutely right. The lease model in #724 addresses the offset bypass in a much cleaner and safer way...

PR #720, opened 2026-08-12
multisig: reject a cosigner key the device already holds

drk1wi:
The only cosigner dedup is in `MultisigWallet.__init__` (`shared/multisig.py:145-149`)
and it is keyed on fingerprint:

```python
        for xfp, deriv, xpub in self.xpubs:
            self.xfp_paths[xfp] = str_to_keypath(xfp, deriv)

        assert len(self.xfp_paths) == self.N, 'dup XFP'
```

`xpub` is unpacked there and never used. Fingerprints are attacker-chosen text in the
config file, so one key under two labels passes that check and `make_redeem_script`
puts it in the script twice. Enrolled through `enrl`, the firmware's own
`yield_addresses()` gives `OP_2 <attacker> <attacker> <victim> OP_3`, two distinct
pubkeys of three. Ceiling is 14-of-15 by one key, 15-of-15 with Skip Checks. The
device refuses to co-sign it at spend time, so the loss is deposits into a wallet
with less redundancy than it displays.

Three changes:

1. `check_xpub` compares `node.pubkey()` against the legs already accepted. Not the
   serialized xpub: `depth`, `parent_fp` and `child_num` are unchecked at depth>1
   (`:822-829`, `:835-845`), libngu re-emits them verbatim, and `make_redeem_script`
   derives from chaincode and pubkey only (`:96-97`). A string compare falls to a
   four-byte edit of `parent_fp`.

2. The same comparison again over the merged list in `ondevice_multisig_create`.
   `check_xpub` only ever sees one accumulator: slot B is appended at `:1768`,
   `list(set(xpubs))` at `:1773` dedups exact tuples only, and both the CCC filter at
   `:1798` and the `add_own_xpub` insert at `:1817` are keyed on xfp.

3. A key the device already holds is refused whatever account it claims. The xfp in a
   `ccxp-*.json` is plain text (`:1540`), so anyone who has received an export can
   relabel it and offer the key back at a different account, where the pubkeys differ
   and no equality test sees it. The file supplies the derivation path, so the device
   derives at that path and compares.

On the CCC path this is worse: `M` is forced to 2, so a planted copy of key A gives
the master seed two of the legs and it meets the threshold alone, without the
spending policy binding.

`Multisig Wallets → Import` is not covered. The same key at a different account still
gets through there, because the cross-check in `check_xpub` (`:856-863`) only runs
when the xfp claims to be the device's own, and covering it means running the derive
on `from_psbt` during signing.

---

### DIFF
--- a/shared/multisig.py
+++ b/shared/multisig.py
@@ -864,7 +864,18 @@ class MultisigWallet(WalletABC):
 
         # serialize xpub w/ BIP-32 standard now.
         # - this has effect of stripping SLIP-132 confusion away
-        xpubs.append((xfp, deriv, chain.serialize_public(node, AF_P2SH)))
+        here = chain.serialize_public(node, AF_P2SH)
+
+        # reject a key we already have, even under a different fingerprint
+        # - dedup in __init__ is keyed on XFP, which is just text in the config file
+        # - compare pubkeys, not the serialization: depth, parent_fp and child_num
+        #   are unchecked at depth>1, so one key has many valid spellings
+        here_pk = node.pubkey()
+        for _, _, prev in xpubs:
+            prev_node = chain.deserialize_node(prev, AF_P2SH)
+            assert here_pk != prev_node.pubkey(), 'dup xpub'
+
+        xpubs.append((xfp, deriv, here))
 
         return (xfp == my_xfp)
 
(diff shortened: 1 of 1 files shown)


doc-hex:
Please add a one-liner to `Next-Changelog.md`

scgbckbone (commented):
Thanks for this report and the careful write-up — the pubkey-comparison approach (and the `parent_fp` respelling rationale for why a string compare isn't enough) is exactly right, and the CCC M=2 variant was a sharp catch.

We've opened an alternative implementation in #725 that builds directly on your analysis. Main differences:

- The dedup + own-key checks live in one classmethod, `MultisigWallet.check_unique_cosigner_keys()`, called from the three enrollment choke points (`from_file`, `import_from_psbt`, `ondevice_multisig_create`) instead of three separate hunks.
- That also covers the `Import` / USB-`enrl` / NFC / descriptor paths for the own-key case — the gap your PR body notes. Those paths run unlocked at enrollment time, so the derive-and-compare is available there; no need to push it to signing time.
- `check_xpub` is untouched, so nothing changes on the signing path (`validate_psbt_xpubs`), and there's no per-leg O(N²) re-deserialization.
- The key-C check runs inline in the CCC block, the one place that secret is in scope.
- One side effect we liked: because the dup check no longer raises inside `check_xpub`, the SD coordinator's `except Exception: continue` can't silently swallow it — an attacked 2-of-3 is now rejected loudly instead of quietly becoming a 2-of-2.
- Simulator tests included: same key under two XFPs, and our own key offered back at a different account under a fake XFP.

Credit for the find stays with you — the ChangeLog entry in #725 thanks you for the report. Curious what you think of the approach.

PR #722, opened 2026-08-12
bugfix: signing TOCTOU on PSRAM-staged PSBT

scgbckbone:
USB host could rewrite the staged PSBT in PSRAM with an aligned upld while the approval screen was up; signing re-reads live PSRAM, so the device signed a transaction different from the one shown. Reproduced on simulator: valid signature over a swapped destination.

Fix:
• hash staged bytes at parse, re-verify right before sign_it() (no await between); mismatch → wipe buffer, fail with "Transaction modified". Covers USB/SD/NFC/vdisk/HSM — all pass  through interact()                                                                                                                                                                      
• re-check gated by a PSRAM write-generation counter, so the common path is one integer compare — no signing slowdown
• upld returns busy while any non-transaction approval flow is active                                                                                                                     
                                                                                                                                                                                             
Reported by FreeZ Agent - thank you

### DIFF
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -29,6 +29,26 @@ from msgsign import sign_message_digest
 TXN_INPUT_OFFSET = 0
 TXN_OUTPUT_OFFSET = MAX_TXN_LEN
 
+def psram_sha256(offset, length):
+    # SHA-256 over a region of PSRAM
+    # - read_at is zero-copy (a view), so no big RAM usage here
+    from glob import PSRAM
+
+    rv = sha256()
+    for pos in range(offset, offset+length, 4096):
+        rv.update(PSRAM.read_at(pos, min(4096, offset+length-pos)))
+
+    return rv.digest()
+
+def psram_wipe(offset, length):
+    # zero-out a region of PSRAM, in chunks
+    # - can zero up to 255 bytes past end: always still inside TXN staging area
+    from glob import PSRAM
+
+    z = bytes(256)
+    for pos in range(offset, offset+length, 256):
+        PSRAM.write(pos, z)
+
 class UserAuthorizedAction:
     active_request = None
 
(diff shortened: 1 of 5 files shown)


doc-hex:
- rename `txn_staging_generation` => `txn_write_count`
- I think there is still a possible race between user approval and the signing, as `self.psbt` contains pointers into (low) PSRAM, not copies of values
- so maybe add another check after signing

scgbckbone (commented):
The "race between approval and signing" point: not possible currently. Yes, self.psbt holds pointers into PSRAM rather than copies  but the firmware is cooperative single-threaded asyncio, so USB/NFC tasks can only run at await points, and the span from the digest re-check through sign_it() and result serialization (done_signing() writes the result before its first await) contains no await at all. No writer can run in that span, so the signed bytes are provably the checked bytes. A post-sign check would compare bytes that cannot have changed. Post-serialization tampering can't redirect funds. 

Non-ALL sighash malleability is handled independently: SIGHASH_NONE is already blocked by default, and PR #704 extends the same default-block to SIGHASH_SINGLE (both remain available via the "Sighash Checks: Warn" setting). The TOCTOU fix is sighash-agnostic either way - signed bytes always equal reviewed bytes. 

I've added a plain assert after signing anyway as a tripwire: it cannot trigger today, but it fails loudly if a future change ever adds an await or a new writer on this path.

shared/auth.py
```diff
@@ -588,6 +628,12 @@ async def interact(self):
         except BaseException as exc:
             return await self.failure("Signing failed late", exc)
 
+        # tripwire: no writer could have run since the re-check above, as
+        # there is no await between it and signing (single-threaded asyncio),
+        # so this cannot trigger today - it fails loudly if a future change
+        # adds an await or a new TXN-region writer on this path
+        assert PSRAM.txn_write_count == self.parsed_write_count
```
doc-hex, line 639: exactly, this comment and assert is just what was needed.

PR #727, opened 2026-08-14
Improve seed generation

scgbckbone:
Supersedes #713.

Key-mash entropy now timestamps raw, pre-debounce GPIO press edges using `utime.ticks_cpu()`, backed by the ARM DWT cycle counter, instead of `utime.ticks_us()`.

At the configured 120 MHz core clock, this provides approximately 8.33 ns CPU-cycle resolution. The Unix simulator retains a `ticks_us()` fallback because its CPU-cycle counter is not meaningful.

The entropy accounting is also updated:

- The first press establishes the timing reference and receives no entropy credit.
- The following 64 inter-press gaps are conservatively credited with two bits each.
- This requires 65 physical mash presses for 128 credited bits.
- Complete timing gaps and key identities are mixed into the hash, but key identity receives no entropy credit.
- Repeating one key is valid, releases are ignored, and users may continue beyond 65 presses.

## Other seed-generation changes

- [Combine STM32 TRNG, SE1 and SE2 entropy](https://github.com/scgbckbone/firmware/commit/59cc8a46).
- [Reseed using the full 256-bit Secure Element entropy digest](https://github.com/scgbckbone/firmware/commit/7a0db4ea).
- [Require a human entropy source and support key mashing, dice rolls and coin flips](https://github.com/scgbckbone/firmware/commit/5a6b30e5).
- [Warn about dice-only seed generation](https://github.com/scgbckbone/firmware/commit/c43badee).
- [Require supplemental entropy for generated Temporary Seeds and CCC key C](https://github.com/scgbckbone/firmware/commit/049e75a1).
- [Replace Yasmarang with the libngu SHA-256 Hash-DRBG](https://github.com/scgbckbone/firmware/commit/3a6dd091).
- [Reduce the libngu firmware footprint and configure external RNG integration](https://github.com/scgbckbone/firmware/commit/88b65dee).

## Mk5 input handling

The [Mk5 keypad rearm fix](https://github.com/scgbckbone/firmware/commit/2a5a7834) rearms mash input immediately after a debounced release is confirmed. This preserves the full release debounce while preventing display or asynchronous processing from unnecessarily delaying acceptance of the next physical press.


scgbckbone (commented):
**Timing-entropy justification for `MIN_MASH_PRESSES = 65`** (full analysis in the attached PDF)

**TL;DR:** 8 Coldcard Q sessions + 10 Mk5 sessions (2,453 inter-press gaps, raw DWT cycle-counter timestamps captured pre-debounce in the hard IRQ) support crediting **2 bits of supplemental entropy per inter-press gap** with zero credit for key identity. Accounting: the first press establishes the timestamp (uncredited), so 65 presses → 64 gaps → 128 credited bits.

- Worst per-run estimate (most-common-value, 99% two-sided upper bound — stricter than NIST SP 800-90B's one-sided 99%): **2.31 bits/gap (Q)**, **2.402 bits/gap (Mk5)**, under deliberately adverse conditions (single key, metronomic rhythm, non-dominant hand).
- Survives stricter simultaneous-confidence checks (Bonferroni across sessions: 2.18 / 2.25 bits) and robustness checks: attacker-chosen projections, conditioning on coarse gap length, and a first-order Markov predictor all fail to beat the baseline.
- Side result: the ~350 ms pre-fix Mk5 accepted rate was delayed rearming, not a physical limit — fixed by the tip commit (rearm after fully debounced release); post-fix runs show ~1.1–2.0× faster acceptance with no entropy loss.
- Caveats: single operator, few units, not a formal SP 800-90B validation. This is *supplemental* entropy — the primary TRNG/secure-element path is unaffected even if mash entropy were zero.

[mash-entropy-assessment.pdf](https://github.com/user-attachments/files/31074842/mash-entropy-assessment.pdf) full methodology, per-run data, and all 2,453 raw measurements.

doc-hex:
**Review performed by OpenAI Codex using GPT-5.6**, focused specifically on key-input API and operational behavior outside mash mode, including state carried from mash back into ordinary UX.

Codex found one issue:

**Q1 mash mode can leak Caps Lock into later UX.**

`FullKeyboard.process_chg_state()` continues processing the normal SHIFT+SYMBOL Caps Lock shortcut while mash mode is active. If those keys overlap during random mashing, `self.caps_lock` is toggled and remains toggled after `stop_mash()`. Later text-entry screens can therefore unexpectedly start in caps mode.

Recommended fix:

```python
if (not self._mash_mode and meta_chg
        and symbol_down and shift_down):
    self.caps_lock = not self.caps_lock
```

A Q1 mash-to-normal transition test covering SHIFT+SYMBOL would catch this.

Otherwise, Codex found no regression when mash mode is not entered:

- `get()` and `get_nowait()` still return key strings.
- `inject()` retains its existing interface.
- `_key_event(key)` remains backward-compatible.
- Only the private `_changes` queue now stores `(key, timestamp)` tuples; repository consumers use the compatibility wrappers.
- Ordinary Q1 scanning is unchanged while `_mash_mode` is false.
- The Mk4/Mk5 membrane refactor preserves the normal press/debounce/release sequence.
- HSM and standard `PressRelease` consumers continue receiving plain key strings.

Codex also checked the post-mash handoff. Apart from the Caps Lock leak, IRQ restoration and queued release handling appear sound.

One coverage gap remains: the new tests exercise mash transitions and simulator injection, but not ordinary physical membrane scanning. A hardware smoke test should cover PIN entry, navigation, held keys, and cancel/accept immediately after leaving mash on Mk4/Mk5.

scgbckbone:
good catch - fixed in https://github.com/Coldcard/firmware/pull/727/commits/640af4dc3782563000ad906cb7a793958644aaa3

PR #729, opened 2026-08-14
bugfix: simulator crash from desynced LED pipe

hitechhayekian:
## Problem

Selecting **Advanced → Danger Zone → Bless Firmware** — or any other path reaching `pa.greenlight_firmware()` — kills the unix simulator immediately:

```
File "unix/./simulator.py", line 1146, in start
    mask, lset = c
ValueError: not enough values to unpack (expected 2, got 1)
```

## Root cause

The LED pipe is framed as two bytes per message, `[mask, state]`. That's what the reader expects (`unix/simulator.py:1140-1148`) and what every writer in `unix/variant/` emits:

| writer | bytes |
| --- | --- |
| `ckcc.py:19` | `b'\xff\x01'` |
| `ckcc.py:92` | `b'\x01\x00'` |
| `ckcc.py:96` | `b'\x01\x01'` |
| `machine.py:39` | `bytes([bm, (bm if n else 0)])` |

`sim_secel.py:205` — the SE1 "greenlight firmware" path (submethod 5) — was still writing a *single* byte, left over from the older 4+4-bit protocol that the reader's own comment refers to.

That one stray byte shifts the stream permanently: every later `read(2)` straddles a message boundary and decodes garbage LED state, and the lone trailing byte then fails to unpack. The crash is immediate rather than delayed because `led_rx` is set `O_NONBLOCK` (`simulator.py:993-996`), so `read(2)` returns the single available byte instead of waiting for a second one.

Decoding the real emitted sequence shows the corruption ahead of the crash: the green (SE1) bit tracks correctly until the stray byte, then flips off and stays off.

## Fix

Two lines in the same three-line block:

1. Write `[mask, state]` like every other caller — `b'\x01\x01'`, matching what `ckcc.py:96` already does to light the same LED.

2. Set `genuine_led` on the module rather than on a local name. `from ckcc import genuine_led` binds a local, so the following `genuine_led = True` rebound that local and left `ckcc`'s module global untouched — greenlighting the firmware never actually marked it genuine, and a later `ckcc.presume_green()` would still assert. `import ckcc` + `ckcc.genuine_led = True` reaches the same global that `ckcc`'s own method 4 writes through `global genuine_led`.

```python
elif submethod == 5:
    # greenlight firmware
    import ckcc
    ckcc.genuine_led = True
    ckcc.led_pipe.write(b'\x01\x01')     # [mask, state] - two bytes always
```

## Verification

Both halves checked against the frozen `ckcc` in a built `coldcard-mpy`, since the local-vs-module distinction is the whole point of the second fix:

```
initial ckcc.genuine_led = True
after clear              = False
OLD form -> ckcc.genuine_led = False | local name = True
OLD form -> presume_green(): AssertionError  <-- the bug
NEW form -> ckcc.genuine_led = True
NEW form -> presume_green(): PASSED
```

## Reproducing

1. Build the unix simulator and run `./simulator.py`
2. Advanced → Danger Zone → **Bless Firmware**

Before: the simulator exits with the traceback above. After: no crash. Nothing visible happens, which is correct — `bless_flash` renders no message, and the green LED is already lit from boot (`ckcc.py:19` writes "all off, except SE1 green"), so setting that bit again is a visual no-op.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scgbckbone (commented):
Thanks for the fix. The two-byte LED pipe change is correct, but the added AST-based tests do not provide useful behavioral coverage and add unnecessary maintenance overhead. Please remove `testing/test_led_pipe.py` and keep the focused simulator fix.

  Also, please import `ckcc` and set `ckcc.genuine_led` = True directly. The current `from ckcc import genuine_led` assignment only changes a local name, leaving the module state unchanged.

doc-hex:
> Thanks for the fix. The two-byte LED pipe change is correct, but the added AST-based tests do not provide useful behavioral coverage and add unnecessary maintenance overhead. Please remove `testing/test_led_pipe.py` and keep the focused simulator fix.

Yes

> Also, please import `ckcc` and set `ckcc.genuine_led` = True directly. The current `from ckcc import genuine_led` assignment only changes a local name, leaving the module state unchanged.

Ahem. That's our bug, but yes ... it's a bug we should fix in same patch.

hitechhayekian:
Both addressed in 33b11dfe.

- Dropped `testing/test_led_pipe.py`.
- `genuine_led` is now set on the module:

```python
elif submethod == 5:
    # greenlight firmware
    import ckcc
    ckcc.genuine_led = True
    ckcc.led_pipe.write(b'\x01\x01')     # [mask, state] - two bytes always
```

I checked the second one against the frozen `ckcc` in a built `coldcard-mpy` rather than just reasoning about it, since the local-vs-module distinction is the whole point:

```
initial ckcc.genuine_led = True
after clear              = False
OLD form -> ckcc.genuine_led = False | local name = True
OLD form -> presume_green(): AssertionError
NEW form -> ckcc.genuine_led = True
NEW form -> presume_green(): PASSED
```

So the old line was costing more than the LED state — a later `ckcc.presume_green()` would still assert after greenlighting.

PR is down to the three lines in `unix/variant/sim_secel.py` now. I left the fixes as a follow-up commit rather than amending; happy to squash to one if you'd prefer that before merge.

doc-hex:
this is great. add a one-liner to Next-Changelog and link to your github profile, and I will merge

hitechhayekian:
Conflict was the 5.6.1 packaging landing while this branch was open — `Next-ChangeLog.md` had been rolled into `ChangeLog.md` and reset. Merged master in and put the one-liner in the fresh Shared section, replacing the `- tbd`:

```
- Bugfix: Simulator crashed on Bless Firmware, due to a desynced LED pipe. Thanks to
  [@hitechhayekian](https://github.com/hitechhayekian).
```

PR is back to `unix/variant/sim_secel.py` (+3/-3) plus that changelog line, and GitHub shows it mergeable again.

PR #730, opened 2026-08-15
bugfix: simulator crashed, then hung, on Destroy Seed

hitechhayekian:
Destroy Seed could not complete in the unix simulator. Two separate problems, one behind the other.

### 1. `pyb.Flash` was missing (crash)

`wipe_flash_filesystem()` drives internal flash as a block device via `pyb.Flash(start=0)` (`shared/files.py:59`), but `unix/variant/pyb.py` never defined `Flash`:

```
File "shared/actions.py", line 609, in clear_seed
File "shared/files.py", line 59, in wipe_flash_filesystem
AttributeError: 'module' object has no attribute 'Flash'
```

This fails *after* both confirmations and after trick PINs are cleared, so the wallet is left half-wiped.

Adds a `Flash` block-device stub alongside the existing `SDCard` one, using the same "pretend to work" approach: 512-byte sectors, as `files.py:61` asserts, and writes that go nowhere. Also adds `testing/test_sim_pyb.py`, which holds the simulator's `pyb` to the attributes `shared/` actually references — that is what would have caught this before it reached a menu action.

### 2. The callgate then hung the simulator

With the crash fixed, Destroy Seed ran to completion and sat on "Clearing..." forever. `ckcc.oneway()` printed a NOTE and then looped on `utime.sleep(60)`, so the last drawn frame stayed on screen — a finished wipe and a crash look identical.

It now raises `SystemExit` and names what the bootloader would have done. All four one-way methods end firmware execution on real hardware (DFU, logout, wipe, brick), so all four stop rather than spin. `machine.py` already does exactly this for `bootloader()`/`soft_reset()`/`reset()`, and the simulator runs with `-i`, so this lands at the REPL the same way a simulated reset always has.

`machine.py`'s `_flush_data()` is deliberately *not* called here: it does `settings.save()`, and `clear_seed()` blanks settings on the line before `fast_wipe()`, so flushing would write back what was just erased.

### Testing

- `testing/test_sim_pyb.py` — 7/7 pass
- Destroy Seed run end to end in the simulator, which now finishes and stops cleanly:

```
NOTE: One-way callgate into bootloader: method=23 arg2=48879
Simulator: would wipe seed & reboot. Exiting.
```

- Nothing under `testing/` references `oneway`, `fast_wipe`, `show_logout`, or `enter_dfu`
- `bare_metal.py` installs its own `my_oneway`, so bare-metal mode is unaffected

Simulator-only — no changes to `shared/` or any on-device code path.

Screenshot 2026-08-14 at 7 20 23 PM

### DIFF
--- a/testing/test_sim_pyb.py
+++ b/testing/test_sim_pyb.py
@@ -1,0 +1,88 @@
+# Regression tests for the simulator's `pyb` stand-in module.
+#
+# unix/variant/pyb.py fakes the pyb module for the unix simulator. Anything the
+# shared firmware reaches for on `pyb` has to exist there, or the simulator dies
+# with AttributeError at the moment that code path runs - which is typically deep
+# in a menu action, long after boot.
+#
+import ast, os, pytest
+
+HERE = os.path.dirname(__file__)
+PYB = os.path.join(HERE, '..', 'unix', 'variant', 'pyb.py')
+SHARED = os.path.join(HERE, '..', 'shared')
+
+# used only on the real device: shared/usb.py guards it with `if is_simulator()`
+GUARDED = {'hid_keyboard'}
+
+def pyb_tree():
+    with open(PYB, 'rt') as fd:
+        return ast.parse(fd.read(), filename='pyb.py')
+
+def sim_pyb_names():
+    # top-level names the simulator's pyb module provides
+    names = set()
+    for node in pyb_tree().body:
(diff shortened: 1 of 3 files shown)


scgbckbone (commented):
Thanks for the contribution. We already cover the Destroy Seed UX in `testing/test_ux.py::test_destroy_seed`. The destructive behavior itself needs to be verified on real hardware, so I don’t think emulating it in the simulator adds enough testing value to justify the additional code.

PR #734, opened 2026-08-16
bugfix: extend signing TOCTOU re-check to TXN output region; cap staged PSBT size

scgbckbone:
Follow-up to [`528677e9`](https://github.com/Coldcard/firmware/commit/528677e9c8efbae4ba44ea2a78ccc71e8ac75da2) ("bugfix: signing TOCTOU on PSRAM-staged PSBT").

## Problem 1: re-check fast path misses the TXN output region

`PSRAMWrapper.write_at()` only bumps `txn_write_count` for writes **below** `MAX_TXN_LEN` (`shared/psram.py`), and the pre-signing digest re-check in `ApproveTransaction.interact()` is gated on that counter (`shared/auth.py`). But one signing flow stages the PSBT at `TXN_OUTPUT_OFFSET` (equal to `MAX_TXN_LEN`, 2MB): the multisig Key Teleport re-sign path. `auth.py` calls `kt_send_psbt(..., psbt_offset=TXN_OUTPUT_OFFSET)` after signing, and the teleport "YOU" entry calls `sign_transaction(..., offset=psbt_offset)` (`shared/teleport.py`).

USB uploads are explicitly permitted while an `ApproveTransaction` is pending (`shared/usb.py`) and may write any offset below `MAX_UPLOAD_LEN` (4MB). A compromised host can therefore rewrite the staged PSBT at or above 2MB **while the second approval screen is up**; the counter never changes, the digest comparison is skipped entirely, and `sign_it()` signs mutated bytes the user never saw—the same bug class as the earlier public fix.

**Fix:** bump the counter for writes into both staging regions (`offset < 2 * MAX_TXN_LEN`). If a write touched the inactive staging region, re-hash the active PSBT and refresh the counter baseline only after proving its bytes are unchanged.

## Problem 2: SD/VirtDisk staging is unbounded

`sign_psbt_file()` used the on-disk size as the PSRAM staging bound (`shared/auth.py`): a 2–4MB binary file overruns the TXN input region into the output region, and larger input can hit an assertion mid-copy. The SD picker already filters at `MAX_TXN_LEN`, but the VirtDisk auto-consume path (`vdisk.py` to `sign_psbt_file`) reaches staging without that filter. Encoded PSBT file size is only an estimate because Base64 and hex input may contain whitespace.

**Fix:** cap the staged decoded data at `MAX_TXN_LEN`. Reject oversized binary files before staging, and reject encoded files only when their actual decoded byte count exceeds the limit.

### DIFF
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -600,13 +600,17 @@ class ApproveTransaction(UserAuthorizedAction):
         # - fast path: nothing wrote to the TXN region since we hashed it,
         #   so there is no need to re-hash in that (common) case
         from glob import PSRAM
-        if (PSRAM.txn_write_count != self.parsed_write_count) and \
-                (psram_sha256(self.offset, self.psbt_len) != self.parsed_sha):
-            # fail closed: wipe the txn, so no signature over modified data
-            psram_wipe(self.offset, self.psbt_len)
-            del self.psbt
-            gc.collect()
-            return await self.failure("Transaction modified")
+        if PSRAM.txn_write_count != self.parsed_write_count:
+            if psram_sha256(self.offset, self.psbt_len) != self.parsed_sha:
+                # fail closed: wipe the txn, so no signature over modified data
+                psram_wipe(self.offset, self.psbt_len)
+                del self.psbt
+                gc.collect()
+                return await self.failure("Transaction modified")
+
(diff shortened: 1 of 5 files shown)


scgbckbone (commented):
OK by KIMI 3

PR #735, opened 2026-08-16
bugfix: bind USB signing to staged PSBT

scgbckbone:
## Summary

- hash upload data only after each successful PSRAM write and reset upload state on errors or protocol aborts
- reject non-contiguous upload offsets and bind stxn length/hash to the staged PSRAM region
- reject host PSBT hashes that differ from the hash computed after parsing, before HSM audit/policy approval
- add simulator regressions for unwritten-block checksum desync, sparse uploads, staged length/hash checks, and HSM hash mismatch


scgbckbone (commented):
OK by KIMI 3

PR #736, opened 2026-08-16
Confirm cloned seed fingerprint

scgbckbone:
Show the restored seed fingerprint for confirmation before Clone Coldcard completes.

This uses the existing backup-restore confirmation and leaves the clone file format and source-device flow unchanged, including compatibility with older source firmware.


doc-hex (approved):

scgbckbone (commented):
OK by KIMI 3

PR #737, opened 2026-08-16
bugfix: enforce CCC velocity block-height floor

scgbckbone:
Legacy CCC velocity policies can persist block_h=0, allowing the chain minimum to be skipped. Enforce the current chain minimum at runtime without rewriting the stored policy.


doc-hex (approved):

scgbckbone (commented):
OK by KIMI 3

PR #739, opened 2026-08-16
spending policy: gate USB restore on related keys

scgbckbone:
## Summary

- require Related Keys before accepting USB backup restore while Spending Policy is active
- preserve temporary-seed restore when Related Keys is enabled


doc-hex (approved):

scgbckbone (commented):
OK by KIMI 3

PR #740, opened 2026-08-17
bugfix: block BIP-85 derivation in Delta Mode

scgbckbone:
Fail closed before BIP-85 derivation reads the master secret in Delta Mode.

This prevents real-master entropy, including reserved duress-wallet indexes, from being derived and exposed.


doc-hex (approved):

scgbckbone (commented):
OK by KIMI 3

PR #741, opened 2026-08-17
bugfix: bound firmware to world checksum

scgbckbone:
Reject firmware past the world-checksum-covered flash boundary and align the signing limit.

Tested: Mk4 and Q1 verifier builds.

### DIFF
--- a/cli/signit.py
+++ b/cli/signit.py
@@ -329,7 +329,7 @@ def doit(keydir, outfn=None, build_dir=None, high_water=False,
     else:
         # new value for Mk4 and later: limited only by final binary size, not SPI flash
         assert FW_MIN_LENGTH <= hdr.firmware_length <= FW_MAX_LENGTH_MK4, hdr.firmware_length
-        USB_MAX_LEN = 1472 * 1024
+        USB_MAX_LEN = FW_MAX_LENGTH_MK4
 
     assert hdr.firmware_length <= USB_MAX_LEN, \
         "too big for our USB upgrades: %d = %d bytes too big" % (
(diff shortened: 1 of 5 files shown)


stm32/mk4-bootloader/verify.c
```diff
@@ -25,6 +25,7 @@
 // 2 megabyte on mk4
 // - less the LFS2 filesystem which contains settings and such that change at runtime normally.
 #define MAIN_FLASH_SIZE          ((2<<20) - (512*1024))
+#define WORLD_CHECKSUM_FW_MAX    (MAIN_FLASH_SIZE - (FIRMWARE_START - BL_FLASH_BASE))
```
doc-hex: maybe FW_MAX_LENGTH_MK4 should be corrected (in sigheader.*) instead?

scgbckbone (commented):
OK by KIMI 3

stm32/mk4-bootloader/verify.c
scgbckbone: Good point—updated. `FW_MAX_LENGTH_MK4` now stops 512 bytes before the filesystem boundary, leaving room for the DFU wrapper. With 4 KiB-aligned firmware, the practical maximum is 1404 KiB instead of 1408 KiB—only 4 KiB less. The separate verifier limit is gone.

doc-hex:
worried this might block use of the `...-factory.dfu` files. based on this from Codex:

>   One behavioral caveat: shared/actions.py applies the limit to the entire DFU container, not just
>   its firmware element. A multi-element development/factory DFU containing firmware plus a
>   bootloader could now be excluded even when its firmware element is valid. Official upgrade DFUs
>   contain only the firmware element, so normal customer upgrades are unaffected.

scgbckbone:
> worried this might block use of the `...-factory.dfu` files. based on this from Codex:
> 
> > One behavioral caveat: shared/actions.py applies the limit to the entire DFU container, not just
> > its firmware element. A multi-element development/factory DFU containing firmware plus a
> > bootloader could now be excluded even when its firmware element is valid. Official upgrade DFUs
> > contain only the firmware element, so normal customer upgrades are unaffected.

safer solution with 2 constants https://github.com/Coldcard/firmware/pull/741/changes/20f9f52739325628db7ca1312c4e34bc41feb54a

PR #742, opened 2026-08-17
Bugfix: reject invalid highwater values

scgbckbone:
Prevent out-of-range highwater timestamps from continuing into downgrade/OTP handling and triggering a bootloader assertion. Applies to both bootloader dispatcher variants.

No tests run (control-flow-only change).


doc-hex (approved):

scgbckbone (commented):
OK by KIMI 3

PR #743, opened 2026-08-17
Bugfix: clear BIP-39 derivation secrets

scgbckbone:
Clear transient BIP-39 mnemonic and PBKDF2 master-secret objects immediately after root-key derivation, including exception paths.


doc-hex:
this is hardening, not bugfix; imho, omit from changelog

scgbckbone (commented):
OK by KIMI 3

PR #744, opened 2026-08-17
bugfix: invalidate dwld leases before Q1 PSRAM staging

scgbckbone:
Clear stale USB `dwld` leases before Q1 reuses PSRAM for Teleport file staging, incoming BBQr data, or outbound BBQr frames. This prevents a host with a previous backup/result lease from reading replacement contents.


doc-hex (approved):

scgbckbone (commented):
OK by KIMI 3

PR #745, opened 2026-08-17
Bugfix: limit multisig BBQr JSON size

scgbckbone:
Reject oversized Q1 multisig coordinator BBQr imports before decoding or JSON parsing. The 1,100-byte limit matches the existing MicroSD multisig-export limit and prevents crafted BBQr payloads from exhausting the MicroPython heap.

### DIFF
--- a/shared/multisig.py
+++ b/shared/multisig.py
@@ -1646,7 +1646,10 @@ async def ms_coordinator_qr(af_str, my_xfp, chain):
     from ux_q1 import QRScannerInteraction, decode_qr_result, QRDecodeExplained
 
     def convertor(got):
-        file_type, _, data = decode_qr_result(got, expect_bbqr=True)
+        file_type, file_size, data = decode_qr_result(got, expect_bbqr=True)
+        if file_size > 1100:
+            raise QRDecodeExplained('Multisig export is too large')
+
         if isinstance(data, bytes):
             # we expect BBQr, but simple QR also possible here
             data = data.decode()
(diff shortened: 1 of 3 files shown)


shared/multisig.py
```diff
@@ -1646,7 +1646,10 @@ async def ms_coordinator_qr(af_str, my_xfp, chain):
     from ux_q1 import QRScannerInteraction, decode_qr_result, QRDecodeExplained
 
     def convertor(got):
-        file_type, _, data = decode_qr_result(got, expect_bbqr=True)
+        file_type, file_size, data = decode_qr_result(got, expect_bbqr=True)
+        if file_size > 1100:
```
scgbckbone, line 1650: This bounds the multisig coordinator convertor, but the generic QR JSON path stays unbounded: `QRScannerInteraction.scan_json` (shared/ux_q1.py, the `file_type, _, data = decode_qr_result(got, expect_bbqr=True)` convertor) still feeds the full BBQr payload — up to the 2 MB PSRAM cap — straight into `json.loads` with no size check. It's reachable on Q1 via Secure Notes → Import → from other Coldcard → QR (shared/notes.py `import_from_other`), i.e. the same crafted-QR heap-exhaustion scenario this PR fixes here; notably the SD branch of that same notes import already caps files at 100000 bytes, so only the QR side is open. The new regression test covers only the coordinator flow. Smallest fix: put the size bound inside `scan_json`'s convertor as well (or in `decode_qr_result` for `'J'`/`'U'` types), so all QR JSON consumers inherit it instead of each caller remembering.

scgbckbone (commented):
The multisig coordinator cap itself looks right (1100 matches the existing MicroSD limit in ms_coordinator_file, and the check runs before any decode/json.loads; final_size is post-decompression so compression can't bypass it). One sibling path remains open, noted inline.

PR #746, opened 2026-08-17
bugfix: reject BIP-21 display control characters

scgbckbone:
Reject non-printable BIP-21 label, message, and lightning values before Q story rendering. Includes scan-flow regression coverage.

### DIFF
--- a/shared/ux_q1.py
+++ b/shared/ux_q1.py
@@ -12,7 +12,7 @@ from decoders import decode_qr_result
 from ubinascii import hexlify as b2a_hex
 from ubinascii import b2a_base64
 
-from utils import problem_file_line, show_single_address
+from utils import problem_file_line, show_single_address, is_printable
 from public_constants import MSG_SIGNING_MAX_LENGTH
 from glob import numpad         # may be None depending on import order, careful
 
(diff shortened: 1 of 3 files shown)


shared/ux_q1.py
```diff
@@ -1133,6 +1133,8 @@ async def ux_visualize_bip21(proto, addr, args):
     for fn in ['label', 'message', 'lightning']:
         if fn in args:
             val = args.pop(fn)
+            if not is_printable(val):
+                val = '(corrupt)'
             msg += '%s%s: %s\n' % (fn[0].upper(), fn[1:], val)
 
     if args:
```
scgbckbone, line 1140: The new `is_printable` gate covers only the three named values, but the leftover-parameter branch just below still interpolates query-parameter *names* raw: `msg += 'And values for: ' + ', '.join(args)`. In `decode_bip21_text` (shared/utils.py) only values are URL-decoded — `args[k] = url_unquote(v)` leaves the key as the raw query substring, and nothing rejects control bytes in it. A byte-mode QR can therefore carry a literal `\n` + `\x02` inside an unknown parameter name (e.g. `bitcoin:<addr>?amount=1&<LF><0x02>tb1q...=1`), producing a story line that starts with `OUT_CTRL_ADDRESS` — which `lcd_display.draw_story` renders as an address-styled line, the same fake-line effect this PR removes for `label`/`message`/`lightning`. The added test only exercises the three named fields, so this path is uncovered. Smallest fix: sanitize the keys in the join, e.g. `', '.join(k if is_printable(k) else '(corrupt)' for k in args)` (or reject control bytes in keys at parse time).

PR #748, opened 2026-08-17
bugfix: bound Delta Mode PIN reconstruction

scgbckbone:
Bound bootloader Delta Mode PIN reconstruction and fail closed on malformed PIN data. Removes the unnecessary terminator write and prevents the reverse walk from moving before the buffer.

This is a defense-in-depth correctness fix, not an exploitable security issue: supported PIN entry and trick-PIN provisioning cannot produce the malformed inputs needed to reach these bounds.

Validation: Mk4 and Q pins.o compile. No Python test applies because the simulator does not execute bootloader C.

### DIFF
--- a/stm32/mk4-bootloader/pins.c
+++ b/stm32/mk4-bootloader/pins.c
@@ -663,28 +663,38 @@ fail:
 
 // apply_pin_delta()
 //
-    static void
-apply_pin_delta(char *pin, int pin_len, uint16_t replacement, char *tmp_pin)
+    static bool
+apply_pin_delta(const char *pin, int pin_len, uint16_t replacement, char *tmp_pin)
 {
     // Starting with provided on pin as a string, change the last few digits
     // given to be replacement value which gives true pin.
     // - encoding: BCD with 0xf for unchanged
-    
+
+    // tmp_pin is exactly MAX_PIN_LEN bytes; reject unsafe copy lengths.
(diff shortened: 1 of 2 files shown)


scgbckbone (commented):
OK by KIMI 3

stm32/mk4-bootloader/pins.c
```diff
@@ -663,28 +663,38 @@ updates_for_good_login(uint8_t digest[32])
...
+    // cursor points one byte past the next character to inspect.
+    int cursor = pin_len;
+    for(int i=0; i<4; i++) {
+        while((cursor > 0) && (tmp_pin[cursor - 1] == '-')) {
+            cursor--;
+        }
 
-    for(int i=0; i<4; i++, p--) {
-        if(*p == '-') p--;
+        // A valid Delta Mode PIN must provide four digits.
```
doc-hex, line 685: untrue

stm32/mk4-bootloader/pins.c
```diff
@@ -663,28 +663,38 @@ updates_for_good_login(uint8_t digest[32])
...
+        if(cursor == 0) return false;
+        cursor--;
 
-        int here = replacement & 0xf;
+        int replacement_digit = replacement & 0xf;
```
doc-hex, line 689: I prefer `here` to `replacement_digit` but I'm not paid by the token.

doc-hex:
- really not convinced this is at all a useful change

PR #750, opened 2026-08-17
bugfix: enable Secure Notes when notes exist but secnap flag missing …

dmonakhov:
Commit  8f86ed1c0ed2 ("deltamode & secure notes and passwords") add new op-in flag "secnap" 
which enabled if secure notes are present, but  older FW has no such flag, so after upgrade/restore
flag will be in it's default state secnap=false, so even if secure notes are present, they will be not visiable.
This is super inconvenient  behavior especially for non tech-savvy users. Let's enable secnap.

### DIFF
--- a/shared/actions.py
+++ b/shared/actions.py
@@ -749,6 +749,12 @@ async def version_migration():
     # version 5.0.6 is installed
     settings.remove_key('vdsk')
 
+    # Secure notes gate moved from 'notes' key presence to 'secnap' opt-in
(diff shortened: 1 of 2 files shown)


scgbckbone (commented):
Thanks for the PR. We don't want to add migration logic for this, either at boot or during backup restore.

The notes are not lost. The user can enable Secure Notes & Passwords under Advanced/Tools, after which the existing notes become visible again. Given the narrow legacy case and straightforward recovery path, we prefer not to add permanent compatibility code.

PR #757, opened 2026-08-18
bugfix: clear PSRAM on fast wipe

scgbckbone:
Clear all PSRAM before a fast-wipe reset. Also clear PSRAM before application SRAM on the non-resetting wipe path.

Tested affected bootloader objects for Mk4/Mk5 and Q1.

### DIFF
--- a/stm32/mk4-bootloader/dispatch.c
+++ b/stm32/mk4-bootloader/dispatch.c
@@ -563,6 +563,9 @@ firewall_dispatch(int method_num, uint8_t *buf_io, int len_in,
                 mcu_key_clear(NULL);
                 oled_show(screen_wiped);
 
+                psram_wipe();
+                wipe_all_sram();
+
                 LOCKUP_FOREVER();
             }
             rv = EPERM;
--- a/stm32/mk4-bootloader/storage.c
+++ b/stm32/mk4-bootloader/storage.c
@@ -23,6 +23,7 @@
 #include "stm32l4xx_hal.h"
 #include "constant_time.h"
 #include "storage.h"
+#include "psram.h"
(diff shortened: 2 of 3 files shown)


scgbckbone:
Exploitation requires more than physical possession: sensitive data must already be staged in PSRAM, power must be maintained through the wipe, and the attacker needs invasive PSRAM-bus access or a separate memory-read primitive. Normal USB/UI access is insufficient, and power removal clears the data. This change closes that post-wipe remanence window.

stm32/mk4-bootloader/dispatch.c
```diff
@@ -563,6 +563,9 @@ firewall_dispatch(int method_num, uint8_t *buf_io, int len_in,
                 mcu_key_clear(NULL);
                 oled_show(screen_wiped);
 
+                wipe_all_sram();
+                psram_wipe();
```
doc-hex, line 566: better wipe SRAM last, our stack is there

stm32/mk4-bootloader/storage.c
```diff
@@ -826,6 +827,9 @@ fast_wipe(void)
     // - and the check value on main seed will be garbage now
     mcu_key_clear(NULL);
 
+    // clear staged application data that would survive the warm reset
+    psram_wipe_lower();
```
doc-hex: why not full wipe? less code

scgbckbone: agreed - Wiping only half saved roughly 75 ms but relied on normal startup to clear the remainder. Using the existing full wipe is simpler and ensures all PSRAM is cleared before reset.

PR #758, opened 2026-08-18
bugfix: require unique multisig names

scgbckbone:
Reject duplicate multisig wallet names at enrollment and storage boundaries. Auto-generated names receive a numeric suffix, while ambiguous legacy names fail closed in HSM policy validation and ownership lookup.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -75,6 +75,7 @@ This lists the new changes that have not yet been published in a normal release.
 - Bugfix: Reject duplicate cosigner keys, and cosigner keys the device already holds,
   during multisig wallet enrollment. Thanks to [@drk1wi](https://github.com/drk1wi) for reporting this.
 - Bugfix: Reject backup files that request excessive password-derivation work.
+- Bugfix: Reject duplicate multisig wallet names.
 - Bugfix: Separate the SE1 check nonce from the PIN digest. Thanks to
   [@instagibbs](https://github.com/instagibbs) for reporting this issue.
 - Bugfix: Clear volatile application data when the seed is wiped.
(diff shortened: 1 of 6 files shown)


scgbckbone (commented):
**OK by KIMI 3** — deep review passed, verified end to end.

Reviewed all three commits in detail:

1. **Require unique names** — every enrollment transport (SD/VDisk file, USB `enrl`, NFC, QR, PSBT xpub import, on-device create, CCC Build 2-of-N) funnels through `confirm_import()` → `has_similar()` → `commit()`, so the new checks cover all boundaries. Branch ordering in `has_similar` is correct (exact-dup → rename-collision → fresh-dup), and the `commit()` assert can't self-trigger since `storage_idx == -1` there. HSM side fails closed correctly: `ApprovalRule.__init__` runs on every policy load, including boot-to-HSM resume, so pre-existing ambiguous policies also fail closed after upgrade.
2. **Generated names** — `make_unique_name` verified against edge cases (bare-name reuse after deletion, max+1 suffixing, non-digit suffixes ignored, result provably unique). Applied at all four generation sites. Two nits, non-blocking: the length `assert` is a hard fail rather than graceful UX (unreachable in practice), and re-importing an *unnamed* config now yields the "Update NAME only" rename offer instead of "All details are the same" because the default name gets suffixed.
3. **Ownership lookup** — fail-closed on ambiguity is correct; `UnknownAddressExplained` renders as "Unknown Address" as the test expects, and injecting legacy duplicates via `settings_set` is the right way to model pre-fix state.

Simulator test results (Mk5 sim, this branch): `test_unique_wallet_names`, `test_named_wallet_search_ambiguous`, `test_named_wallet_search_fail` (control), `test_overflow[3]`, `test_overflow[15]` — all pass.

Note: restore paths (backup/clone/teleport) write the multisig list wholesale and can't create duplicates, only carry them; with HSM and ownership both failing closed, skipping restore-time dedupe is acceptable.

PR #760, opened 2026-08-19
testing: core v32 compat PSBTv2

scgbckbone:


scgbckbone:
new version of core (v32+) will return PSBTv2 by default - some of our test need to cope with that

PR #761, opened 2026-08-19
Make multisig rename explicit

scgbckbone:
Adds an explicit Rename action to the multisig wallet menu. Reimporting an identical wallet is treated as a duplicate instead of implicitly renaming it.

Tests cover successful rename, name collisions, cancellation, and affected CCC/signing flows.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -76,6 +76,9 @@ This lists the new changes that have not yet been published in a normal release.
   during multisig wallet enrollment. Thanks to [@drk1wi](https://github.com/drk1wi) for reporting this.
 - Bugfix: Reject backup files that request excessive password-derivation work.
 - Bugfix: Reject duplicate multisig wallet names.
+- Change: Multisig wallet names can now be changed with a dedicated `Rename`
+  action in the wallet menu. Reimporting an enrollment file or descriptor no
+  longer renames an existing wallet.
 - Bugfix: Separate the SE1 check nonce from the PIN digest. Thanks to
   [@instagibbs](https://github.com/instagibbs) for reporting this issue.
 - Bugfix: Clear volatile application data when the seed is wiped.
--- a/shared/multisig.py
+++ b/shared/multisig.py
@@ -422,10 +422,7 @@ class MultisigWallet(WalletABC):
 
     def has_similar(self):
         # check if we already have a saved duplicate to this proposed wallet
-        # - return (name_change, diff_items, count_similar) where:
-        #   - name_change is existing wallet that has exact match, different name
-        #   - diff_items: text list of similarity/differences
-        #   - count_similar: same N, same xfp+paths
+        # - returns (is_duplicate, differences)
 
         lst = self.get_xfp_paths()
         c = self.find_match(self.M, self.N, lst, addr_fmts=[self.addr_fmt])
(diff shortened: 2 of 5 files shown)


scgbckbone:
The bug was originally fixed in an earlier implementation by tracking whether a wallet name was supplied by the user or generated automatically, then using that information during duplicate detection.

This PR fixes the same bug by removing implicit rename-on-import entirely. Importing now either creates a new wallet or reports a duplicate; renaming is a separate, explicit menu action.

Both approaches are covered by the same `test_reimport_unnamed_descriptor` regression test:

1. Import an unnamed descriptor → create `2-of-3`.
2. Reimport the exact same descriptor → report `Duplicate wallet`.
3. Import a different unnamed 2-of-3 descriptor → create `2-of-3 #2`.

The bug occurs because automatic naming happens before duplicate detection. Reimporting the same unnamed descriptor may produce the candidate name `2-of-3 #2`, which the old code interprets as a request to rename the existing wallet—even though the user supplied no name and only replayed the same descriptor.

The original fix needs to preserve and check how the candidate name originated. This PR makes that distinction unnecessary: duplicate detection depends only on wallet identity, while name changes happen exclusively through the Rename action.

scgbckbone (commented):
## Review: "Make multisig rename explicit"

**Verdict: Safe to merge.** I reviewed every line, traced all callers, checked downgrade/old-firmware interaction, and ran the affected tests against a simulator built with the PR code.

### What it does

- `has_similar()` (`shared/multisig.py:423`) now returns `(is_duplicate, differences)` instead of `(name_change_wallet, diff_items, count)`. The implicit "reimport same wallet under a new name = rename" path is gone; reimporting identical content is now always a duplicate.
- New explicit `Rename` menu item + `ms_wallet_rename()` (`shared/multisig.py:1486`) in the per-wallet menu.
- Changelog updated; tests updated/added.

### Checks performed

- **Branch-by-branch behavior matrix** of `has_similar`: every outcome (xpubs-order mismatch warning, BIP-67 clash, key-order block, name collision, similar-wallet warning, clean import) behaves identically to master except the intended removal of import-as-rename. The tricky `isinstance(diff_items, list)` distinction (list=block, set=warning) is correctly preserved by the new boolean.
- **All import choke points** — NFC/QR/SD/USB enroll (`NewEnrollRequest` → `confirm_import`), PSBT-embedded enrollment (`psbt.py:1395`), CCC — all funnel through the modified `confirm_import`; nothing else calls `has_similar`.
- **New rename UX**: `name_is_used(name, ms.storage_idx)` correctly excludes self; `ux_input_text` on both Mk4 (chars 32–126) and Q1 (`ch_remap` 32–126) constrains input to the same ASCII/1–20 envelope that import validation enforces, so renamed names can't break exports or older firmware; the pop + `update_contents()` pattern matches `ms_wallet_delete` exactly.
- **Old firmware / downgrade**: storage format untouched (only the name string changes), index-preserving rename is actually *safer* than the old delete-and-re-append (which moved `storage_idx` and could lose the wallet if `commit()` raised out-of-space after `delete()`). Nothing a downgraded device reads differently.
- **Ran the tests on a simulator built with the PR** (this repo has no CI): all 8 touched/new tests in `test_multisig.py` pass on Mk5, plus `test_rename_wallet`/`test_reimport_unnamed_descriptor`/`test_unique_wallet_names` on Q1, and `test_sign.py::test_low_R_grinding`.

### Findings (minor, non-blocking)

1. **Message detail loss**: the "same content, different name" and "same content, name used elsewhere" cases now both show a bare `Duplicate wallet.` instead of `All details are the same as existing!` / `name already exists`. Cosmetic only.
2. **`test_ccc.py::test_multiple_multisig_wallets` unverified locally** — needs `bitcoind`, which isn't installed in this environment. Its changes are test-only, and the firmware behavior it covers (rename via CCC menu) is exercised by `test_rename_wallet`. Worth one bench run with bitcoind before release.
3. **`MultisigOutOfSpace` from rename `commit()` is uncaught** — user would see an error screen instead of a clean story; practically impossible (name length bounded by 20) and still better than the old flow's delete-then-commit data-loss window.

No blockers. Safe to release, with the CCC test run on a bitcoind-equipped machine as the one remaining checkbox.

OK by KIMI 3

scgbckbone:
Fixed nit issue 1 from the KIMI 3 review above: duplicate-wallet messages now preserve the relevant detail instead of showing only `Duplicate wallet.`

- `Duplicate wallet. All details are the same as existing!`
- `Duplicate wallet. Name already exists.`

PR #763, opened 2026-08-20
bugfix: integer underflow in PSRAM virtual-disk file import (OOB write from hostile FAT metadata)

Amiga500:
## Summary

`psram_copy_file()` and `psram_mmap_file()` (`stm32/COLDCARD_MK4/psramdisk.c`, shared by Q1 via symlink) trust FAT metadata of the PSRAM virtual disk, which is exposed over USB Mass Storage and therefore fully writable by a compromised USB host.

The final fragment length is computed as:

```c
len = actual_len - so_far;   // uint32_t arithmetic
```

where `actual_len = fp.obj.objsize` (directory entry size field) and `so_far` is the sum of the bytes covered by previous cluster-chain fragments. FatFs' `CREATE_LINKMAP` follows the cluster chain to EOC **without validating it against `objsize`**, so a crafted FAT (chain covering more bytes than `objsize`, e.g. on a `.dfu`/`.psbt` file dropped by the host) makes `so_far > actual_len` and `len` underflows to ~4 GB. The subsequent `memcpy(dest, spot, len)` then writes attacker-influenced data out of bounds:

- the pre-write check `dest + actual_len + 3 >= PSRAM_TOP_BASE` is insufficient and can itself wrap around for `objsize` near 4 GB;
- the `dest >= PSRAM_TOP_BASE` check happened **after** the `memcpy` — too late.

Realistic impact: corruption of everything staged in PSRAM (PSBT/firmware staging areas) and device crash via BusFault. SRAM and flash sit below `0x90000000`, so the forward-only copy does not reach them; no credible path to code execution was identified (hence moderate severity), but it is a deterministic host-triggered memory corruption reachable without user confirmation when Virtual Disk is in "Enable & Auto" mode.

## Fix

- Reject cluster chains whose coverage exceeds the file size (`so_far > actual_len`, and per-fragment `len > remaining`) before the subtraction — no more underflow.
- Verify destination bounds with 64-bit arithmetic **before** each `memcpy`, eliminating both the post-hoc check and the wrap-around.
- Same guard applied to `psram_mmap_file()`, which hands attacker-influenced (address, length) tuples to MicroPython.

## Verification

Logic validated with a standalone host harness replicating the fixed loop: legit single/multi-fragment files pass; hostile chains (oversized coverage, `objsize` near 4 GB, degenerate zero-size) are all rejected. A full firmware build was not possible in the authoring environment (no ARM toolchain / uninitialized submodules) — a run through `repro-build.sh` before merge would be appreciated.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -16,6 +16,9 @@ This lists the new changes that have not yet been published in a normal release.
 - Bugfix: Simulator crashed on Bless Firmware, due to a desynced LED pipe. Thanks to
   [@hitechhayekian](https://github.com/hitechhayekian).
 - Bugfix: With an empty master wallet and an active temporary seed, keep imports and backup restores temporary instead of treating them as master-seed changes.
+- Bugfix: Reject PSRAM virtual-disk files whose FAT cluster chain covers more bytes than the
+  file size, fixing an integer underflow in `psram_copy_file`/`psram_mmap_file` that allowed
+  out-of-bounds PSRAM writes from a compromised USB host.
 
 # Mk Specific Changes
 
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -17,8 +17,9 @@ This lists the new changes that have not yet been published in a normal release.
   [@hitechhayekian](https://github.com/hitechhayekian).
 - Bugfix: With an empty master wallet and an active temporary seed, keep imports and backup restores temporary instead of treating them as master-seed changes.
 - Bugfix: Reject PSRAM virtual-disk files whose FAT metadata is inconsistent with the
-  declared file size (oversized cluster chains, oversized fragment counts, or final
-  remainders exceeding the final fragment's capacity), fixing an integer underflow in
+  declared file size (oversized cluster chains, oversized fragment counts, spurious
+  trailing fragments, final remainders exceeding the final fragment's capacity, or
+  filesystems with more than one sector per cluster), fixing an integer underflow in
   `psram_copy_file`/`psram_mmap_file` that allowed out-of-bounds PSRAM writes, reads,
   and mappings from a compromised USB host.
 
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -16,9 +16,11 @@ This lists the new changes that have not yet been published in a normal release.
 - Bugfix: Simulator crashed on Bless Firmware, due to a desynced LED pipe. Thanks to
   [@hitechhayekian](https://github.com/hitechhayekian).
 - Bugfix: With an empty master wallet and an active temporary seed, keep imports and backup restores temporary instead of treating them as master-seed changes.
-- Bugfix: Reject PSRAM virtual-disk files whose FAT cluster chain covers more bytes than the
-  file size, fixing an integer underflow in `psram_copy_file`/`psram_mmap_file` that allowed
-  out-of-bounds PSRAM writes from a compromised USB host.
+- Bugfix: Reject PSRAM virtual-disk files whose FAT metadata is inconsistent with the
+  declared file size (oversized cluster chains, oversized fragment counts, or final
+  remainders exceeding the final fragment's capacity), fixing an integer underflow in
+  `psram_copy_file`/`psram_mmap_file` that allowed out-of-bounds PSRAM writes, reads,
+  and mappings from a compromised USB host.
 
 # Mk Specific Changes
 
--- a/stm32/COLDCARD_MK4/psramdisk.c
+++ b/stm32/COLDCARD_MK4/psramdisk.c
@@ -607,6 +607,13 @@ mp_obj_t psram_mmap_file(mp_obj_t unused_self, mp_obj_t fname_in)
         mp_raise_ValueError(MP_ERROR_TEXT("unmountable"));
     }
 
+    // The USB host can rewrite the BPB, so a mountable filesystem might
+    // use more than one sector per cluster; the fragment math below works
+    // on 512-byte sectors, so refuse anything else.
+    if(vfs.fatfs.csize != 1) {
+        mp_raise_ValueError(MP_ERROR_TEXT("bad clus size"));
+    }
+
     // open the file directly
     FIL fp = {0};
     if(f_open(&vfs.fatfs, &fp, fname, FA_READ) != FR_OK) {
@@ -665,7 +672,10 @@ mp_obj_t psram_mmap_file(mp_obj_t unused_self, mp_obj_t fname_in)
             // final cluster might include some bytes past the EOF, but the
             // declared remainder must fit in this fragment's real capacity
             len = fp.obj.objsize - so_far;
-            if(len > capacity) {
+            if((len < 1) || (len > capacity)) {
+                // len==0: earlier fragments already cover the whole file, so
+                // this trailing fragment (e.g. chain of a zero-size file) is
+                // oversized metadata; reject it
                 mp_raise_ValueError(MP_ERROR_TEXT("bad file size"));
             }
         } else {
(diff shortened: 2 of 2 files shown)


scgbckbone (commented):
are you planning to re-open ?

Amiga500:
I have sent an email 6 hours ago to security@coinkite.com with all info

Amiga500:
but i can reopen it right away without any problems if you give me the ok

doc-hex:
Ok, to reopen this PR... it's very not hidden at this point.

This fix can be in the next release as it didn't make today's cutoff.

stm32/COLDCARD_MK4/psramdisk.c
```diff
@@ -648,10 +648,19 @@ mp_obj_t psram_mmap_file(mp_obj_t unused_self, mp_obj_t fname_in)
         }
         uint32_t len = num_clusters*BLOCK_SIZE;
 
+        // hostile/corrupt FAT metadata could describe a cluster chain that
+        // covers more bytes than the file size; refuse the underflow below
+        if(so_far > fp.obj.objsize) {
```
scgbckbone, line 667: The raw fragment count needs validation before this point. `mapping[]` stores a `DWORD`, but `block_to_ptr()` accepts `uint16_t num_blk`; for example, 65,536 narrows to zero before the call's range check. A start block near the disk end can then pass, and the final branch can expose or read beyond PSRAM. Read the count as `uint32_t` and reject values outside `1..BLOCK_COUNT` before calling `block_to_ptr()` here and in `psram_copy_file()`.

stm32/COLDCARD_MK4/psramdisk.c
```diff
@@ -732,19 +743,31 @@ mp_obj_t psram_copy_file(mp_obj_t unused_self, mp_obj_t offset_in, mp_obj_t fnam
...
+            mp_raise_ValueError(MP_ERROR_TEXT("bad file size"));
+        }
+
         if(i == num_used-1) {
             // final cluster might include some bytes past the EOF
             len = actual_len - so_far;
             // align4
             len = (len + 3) & ~0x3;
         } else {
+            if(len > actual_len - so_far) {
```
scgbckbone, line 800: This capacity check only covers non-final fragments. A directory entry can declare a final remainder larger than the final link-map fragment (for example, a 1,024-byte file backed by one 512-byte cluster at the last disk block); the final branch then copies or maps past PSRAM. Preserve the fragment capacity (`num_clusters * BLOCK_SIZE`) and reject a final remainder larger than it in both loops, before the copy path's alignment.

scgbckbone (changes requested):
Two hostile FAT/link-map shapes can still bypass these guards and produce out-of-bounds PSRAM reads or mappings. The fragment count must be validated before it is narrowed by `block_to_ptr()`, and the declared final remainder must be checked against the final fragment's physical capacity.

stm32/COLDCARD_MK4/psramdisk.c
Amiga500: Fixed in cf3fe229. `num_clusters` is now read as `uint32_t` and rejected outside `1..BLOCK_COUNT` before the `block_to_ptr()` call, in both `psram_mmap_file()` and `psram_copy_file()`, so the 65536-to-0 narrowing can no longer bypass the range check.

stm32/COLDCARD_MK4/psramdisk.c
Amiga500: Fixed in cf3fe229. The fragment capacity (`num_clusters * BLOCK_SIZE`) is now preserved and the final remainder (`objsize - so_far`) is rejected when it exceeds that capacity, before the align-4 in the copy path — applied to both loops. So a 1024-byte entry backed by a single 512-byte final cluster is refused instead of copying past the fragment.

scgbckbone:
ACK cf3fe22

nit: the changelog only mentions oversized chains and OOB writes while the patch also prevents undersized final fragments and OOB reads/mappings - please update that wording and resolve the changelog conflict with master

otherwise LGTM

stm32/COLDCARD_MK4/psramdisk.c
Amiga500: Already fixed in cf3fe229 (see reply above); changelog wording updated in dec0158e to also mention oversized fragment counts. Resolving.

stm32/COLDCARD_MK4/psramdisk.c
Amiga500: Already fixed in cf3fe229 (see reply above); changelog wording updated in dec0158e to mention undersized final fragments and OOB reads/mappings. Resolving.

doc-hex:
Two findings from reviewing the current patch:

1. **Fragment capacity assumes one sector per cluster.** `CREATE_LINKMAP` reports run lengths in clusters, but `num_clusters * BLOCK_SIZE` and `block_to_ptr(..., num_clusters)` treat the count as sectors. Because the USB host can rewrite the BPB, FatFs can mount a valid filesystem with `csize > 1`; such files can then be rejected or, for fragmented/overallocated layouts, copied or mapped with skipped data. Please either compute a checked sector count as `num_clusters * vfs.fatfs.csize` and use it for both `block_to_ptr()` and capacity, or explicitly reject `csize != 1` after mounting.

2. **One trailing fragment is still accepted when preceding fragments exactly cover the declared size.** If `so_far == objsize`, the final branch produces `len == 0` and accepts the oversized chain; `mmap()` consequently returns a spurious zero-length tuple. Please reject this case when another fragment remains (including a zero-size file with an allocated chain).

Non-blocking follow-ups: the PR currently conflicts with `master` in `Next-ChangeLog.md`, and the hostile-FAT cases are only covered by an external harness rather than repository regression tests.

Amiga500:
Thanks for the careful re-review. Both findings addressed in 457acd8d:

1. Sectors vs. clusters (csize > 1) — I went with the explicit-reject option: right after f_mount(), both psram_mmap_file() and psram_copy_file() now refuse any mounted filesystem with vfs.fatfs.csize != 1 (ValueError("bad clus size")). I preferred this over computing a checked sector count (num_clusters * csize) because the whole downstream math — clst2sect()'s scaling, block_to_ptr()'s sector range check, and the capacity bookkeeping — is built around 512-byte sectors; pinning csize == 1 keeps one invariant instead of threading a second unit through every calculation. A host-rewritten BPB now just makes the file unreadable instead of mis-mapped.

2. Trailing fragment after exact coverage — the final branch now rejects len < 1 as well as len > capacity:

len = fp.obj.objsize - so_far;
if((len < 1) || (len > capacity)) {
    mp_raise_ValueError(MP_ERROR_TEXT("bad file size"));
}

So when preceding fragments already cover objsize exactly (so_far == objsize), a remaining fragment is rejected as oversized-chain metadata instead of producing a spurious zero-length tuple from mmap(). This also covers the zero-size-file-with-allocated-chain case you mentioned (objsize == 0 with any link map).

Follow-ups:

    The Next-ChangeLog.md conflict with master is resolved in the current head (both entries kept), and the entry now also mentions spurious trailing fragments and multi-sector clusters.
    On the regression tests: agreed that repo-side coverage would be better than my external harness. The catch is these paths run on bare-metal PSRAM/FAT code with no simulator hook today, so a pytest-style regression would need either a host-compiled harness of psramdisk.c or new simulator plumbing. Happy to add the host-harness variant under testing/ if you think that's worth the build plumbing — let me know your preference.

Re-validated with the standalone harness extended for the new shapes: legit single/multi-fragment files (including normal over-allocation of the final cluster) pass; hostile shapes (csize 2/4, num_clusters u16 wrap, oversized chains, exact-coverage trailing fragment, zero-size file with chain, clusters past end of disk) are all rejected. CodeQL is clean on the diff.

doc-hex:
Move commit 457acd8d onto the branch backing the PR. The fixes currently exist only on copilot/amiga500-fix-psramdisk-fat-underflow, while PR 763 tracks amiga500-fix-psramdisk-fat-underflow.

Amiga500:
@doc-hex Done — the fixes from 457acd8d are now on the PR branch as be54bf8e (cherry-picked onto the PR head; tree verified identical). Ready for re-review.

releases/Next-ChangeLog.md
```diff
@@ -4,7 +4,14 @@ This lists the new changes that have not yet been published in a normal release.
 
 # Shared Improvements - Both Mk and Q
 
-- tbd
+- Bugfix: Simulator crashed on Bless Firmware, due to a desynced LED pipe. Thanks to
```
scgbckbone, line 16: remove

PR #766, opened 2026-08-25
Show TRNG words for entropy verification

scgbckbone:
Restores the pre-5.6.1/1.5.1Q ability to view the device-generated seed before adding user entropy.

Adds **View TRNG Words** to show the full 256-bit STM32 TRNG + SE1 + SE2 seed as 24 BIP39 words for independent verification of dice-roll or coin-flip mixing.

Mixing behavior is unchanged and always uses all 256 bits, including for 12-word wallets.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -4,6 +4,11 @@ This lists the new changes that have not yet been published in a normal release.
 
 # Shared Improvements - Both Mk and Q
 
+- Bugfix: Restore the ability to view the device-generated seed before adding user
+  entropy, which was available in the previous dice-roll workflow but was inadvertently
+  removed in 5.6.1/1.5.1Q. The new **View TRNG Words** menu item displays the full
+  256-bit seed from the STM32 TRNG, SE1, and SE2 as 24 BIP39 words, allowing independent
+  verification of dice-roll or coin-flip mixing.
 - Bugfix: Simulator crashed on Bless Firmware, due to a desynced LED pipe. Thanks to
   [@hitechhayekian](https://github.com/hitechhayekian).
 
--- a/shared/seed.py
+++ b/shared/seed.py
@@ -789,27 +789,52 @@ async def collect_mash_entropy():
 
     return md.digest()
 
+async def view_trng_words(_menu, _idx, item):
+    from ux import ux_render_words
+
+    words = bip39.b2a_words(item.arg).split(' ')
+    msg = ('These 24 words encode the full 256-bit device seed (STM32 TRNG + SE1 + SE2) '
+           'before user entropy is mixed in.\n\nAll 256 bits are used, even for a 12-word wallet.'
+           '\n\nUse them to verify dice-roll or coin-flip mixing.\n\nKEEP SECRET: Anyone with '
+           'these words and your complete user entropy can recreate your wallet seed.'
+           '\n\n%s' % ux_render_words(words))
+    try:
+        await ux_show_story(msg, title='TRNG Words', sensitive=True)
+    finally:
+        blank_object(msg)
+
+
+async def pick_user_entropy(base_seed):
+    picked = []
+
+    async def selected(_menu, _idx, item):
+        picked.append(item)
+        the_ux.pop()
+
+    menu = MenuSystem([
+        MenuItem('Mash Keys', f=selected, arg=METHOD_MASH),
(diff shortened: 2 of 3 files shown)


doc-hex:
I don't like this. It's a security risk to show the value, and users cannot independently verify anything with the value shown (correct me if I'm wrong). In the "Dice Rolls" case, showing the value allows full off-device verification, but there is no way to support that with our new approach to entropy.

scgbckbone:
we showed the value in all prior versions - removed in 5.6.1 and 1.5.1Q only. 

It is the only way to make additional dice/coins rolls/flips verifiable. It seems pointless wrt TRNG because you cannot verify how that was created, BUT with the TRNG seed available you can verify if dices/coins are added properly. Check the added test - they do exactly that.

 This is useful and I found that our users were actually doing it . Some contacted me after we removed in previous release.

doc-hex:
Sorry I forgot that flips and rolls are verifiable... so we need this.

scgbckbone:
## Review — "Show TRNG words for entropy verification"

**What it does:** adds a `View TRNG Words` item to the entropy-picker menu in New Seed Words, showing the 256-bit device seed (STM32 TRNG + SE1 + SE2) as 24 BIP39 words *before* user entropy is mixed in. Mixing math and menu flow are otherwise unchanged. Restores visibility lost in 5.6.1/1.5.1Q.

### Verification performed
- Applied the diff onto local master — applies cleanly.
- Ran it on the Mk4 headless simulator (`testing/ENV` venv, `PYSECP_SO` pointing at a local libsecp256k1):
  - 4/4 new tests pass: `test_view_trng_words_verifies_dice_mix[12,24]`, `test_view_trng_words_verifies_coin_mix[12,24]`
  - 21/21 touched existing tests pass: `test_new_wallet` grid, `test_new_wallet_entropy_cancel`, `test_new_wallet_rejects_biased_dice/coin` — no regression.
- Static checks against the codebase:
  - The push/interact/pop idiom in `pick_user_entropy` matches the established pattern (`shared/actions.py:1870-1888`); the callback signature `f(menu, idx, item)` is correct per `MenuItem.activate` (`shared/menu.py:87`). Loop exit works via `the_ux.pop()` making `interact()`'s `top_of_stack() == self` guard false (`shared/menu.py:396-407`).
  - CANCEL item and CANCEL key both land the caller on `picked is None or picked.arg is None` → clean abort.
  - Hygiene: `msg` is blanked in a `finally` (`blank_object` handles str, `shared/stash.py:25`), and `base_seed` is blanked in the caller's `finally` — note this is the same object referenced as the menu item's `arg`, so the menu reference doesn't defeat blanking.
  - The test's hardcoded mix domains match the constants: `b'CC\x01SMD'` = `DOMAIN_SEED + PURPOSE_MASTER + METHOD_DICE` (`shared/seed.py:49-53`), and the 16-byte truncation for 12-word wallets matches `approve_word_list` (`shared/seed.py:894-895`). Test vectors respect the bias guards (9/50 ≤ 0.30 rolls; 64/128 ≤ 0.65 flips).

### Minor findings (non-blocking)
1. **Hardcoded domain bytes in tests** — the tests restate `b'CC\x01SMD'` / `b'CC\x01SMC'` literally rather than composing them from the constants. That's fine as spec-locking, but a one-line comment noting the composition would make future failures less cryptic.
2. **Abort-path nuance** — the old loop used `finally: the_ux.pop()` around `wait_choice()`; the new push+interact flow leaves the picker menu on the UX stack if an `AbortInteraction` propagates out (the main loop then re-interacts with it). Same as other pre-existing menus, harmless, but worth knowing it's a behavioral difference on the abort path.
3. **Story wording** — the secrecy warning is accurate ("these words + your complete user entropy"), but users may mistake the TRNG words for a backup. Adding an explicit "these words are not a wallet backup" line would close that foot-gun.

### Security assessment
Exposure is opt-in mid-ceremony, rendered with `sensitive=True` (flags display-layer side-channel defenses), and blanked afterward. The claim "all 256 bits are used, even for a 12-word wallet" is correct and asserted by the tests. Showing the base seed doesn't weaken the scheme — it enables exactly the independent mixing verification the PR states.

**Verdict: LGTM.** Solid implementation, genuinely-tested (I reproduced the test run locally), and the changelog entry matches the behavior. The nits above are optional.

---
*This review was produced with KIMI 3.*

PR #772, opened 2026-09-01
EDGE testing: stabilize multiprocess state checks

scgbckbone:
## Summary

- wait for the Q1 multisig rename to be committed before reading settings
- reset the simulator seed before constructing the Teleport multisig wallet


scgbckbone (commented):
OK by KIMI 3

PR #773, opened 2026-09-01
bugfix: reject foreign inputs from BIP322 proofs of reserves

scgbckbone:
## Summary

- reject UTXO-bearing foreign inputs from BIP322 Proof of Reserves transactions
- update the existing missing-key-path coverage to require a hard failure
- document the bugfix in `releases/Next-ChangeLog.md`

### DIFF
--- a/shared/psbt.py
+++ b/shared/psbt.py
@@ -1885,7 +1885,11 @@ class psbtObject(psbtProxy):
             # - also finds appropriate multisig wallet to be used
             inp.determine_my_signing_key(i, utxo, self.my_xfp, self, cosign_xfp)
 
-            if self.por322 and i and not inp.num_our_keys:
+            if self.por322 and i and not inp.required_key:
+                # every POR input past to_spend must be one we will actually sign;
+                # num_our_keys is not enough: forged keypaths (incl. zero-xfp
+                # placeholder) plus a partial sig would make a foreign input
+                # look ours and already-signed (required_key is None)
                 foreign_por = True
 
             if inp.required_key and not inp.is_segwit and not inp.utxo:
--- a/shared/psbt.py
+++ b/shared/psbt.py
@@ -1849,6 +1849,7 @@ class psbtObject(psbtProxy):
         total_in = 0
         from_wif_store = []
         prevouts = set()
+        foreign_por = False
 
         for i, txi in self.input_iter():
             # check for duplicate inputs
@@ -1884,6 +1885,9 @@ class psbtObject(psbtProxy):
             # - also finds appropriate multisig wallet to be used
             inp.determine_my_signing_key(i, utxo, self.my_xfp, self, cosign_xfp)
 
+            if self.por322 and i and not inp.num_our_keys:
+                foreign_por = True
+
             if inp.required_key and not inp.is_segwit and not inp.utxo:
                 raise FatalPSBTIssue('Legacy input #%d requires non-witness UTXO' % i)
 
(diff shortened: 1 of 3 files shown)


shared/psbt.py
```diff
@@ -1884,6 +1885,9 @@ def consider_inputs(self, cosign_xfp=None):
             # - also finds appropriate multisig wallet to be used
             inp.determine_my_signing_key(i, utxo, self.my_xfp, self, cosign_xfp)
 
+            if self.por322 and i and not inp.num_our_keys:
```
scgbckbone: **Major: this gate is bypassable because `num_our_keys` is attacker-controllable.**

`parse_subpaths()` (psbt.py:315) increments `num_our_keys` for any keypath whose xfp equals `my_xfp` — and the zero-xfp placeholder feature (psbt.py:306-307) rewrites `here[0] == 0` to `my_xfp` on the fly, so the attacker doesn't even need to know the victim's fingerprint. Attack PSBT: input 0 = valid `to_spend` for the victim's address; inputs 1..N = foreign UTXOs each carrying `witness_utxo` + keypath `{foreign_pk: [0, ...]}` + a partial signature (garbage bytes suffice — existing `part_sigs` are only counted at psbt.py:679, never verified).

Result: `num_our_keys == 1`, so this check passes; and `fully_signed` becomes True, so `determine_my_signing_key()` early-returns with `required_key = None` (psbt.py:801) — nothing ever attempts to sign the input, so the "path led to wrong pubkey" failure at signing time is never reached either. The only fallout is two non-blocking UX warnings ("Partly Signed Already", "Zero XFP"). The device signs input 0 with SIGHASH_ALL — committing to all foreign prevouts — and exports a BIP322 proof bundling foreign UTXOs with the victim's: the exact bug class this PR targets, needing only two extra PSBT fields versus the variant being fixed.

Suggest gating on whether we will actually sign the input, rather than on keypath metadata:

```python
if self.por322 and i and not inp.required_key:
    foreign_por = True
```

I verified this doesn't break legitimate flows: existing POR tests use unsigned own inputs (`required_key` set), WIF-store POR inputs still resolve `required_key`, and the multisig POR test preloads only M-1 placeholder cosigner sigs so `fully_signed` stays False. The only newly-rejected case is a POR containing already-fully-signed inputs — arguably correct to reject for proof-of-reserves semantics.

scgbckbone (commented):
Deep review of the fix and its surroundings (`consider_inputs`, `parse_subpaths`, `determine_my_signing_key`, the signing loop, and the HSM/teleport POR callers).

The direction is right and the change is well-placed: the missing-keypath variant of foreign POR inputs is now hard-failed instead of passing with a "Limited Signing" warning, the ordering is correct in all callers (`validate()` sets `por322` before `consider_inputs` in both auth.py and teleport.py, HSM approval runs after), WIF-store inputs are not falsely rejected (they get `num_our_keys = 1` in `determine_my_signing_key`), and the updated test genuinely flips from approval-UX to `Failure` when the fix is present.

However, the new gate is bypassable with a slightly more sophisticated PSBT — see the inline comment. The check keys on `num_our_keys`, which is derived purely from attacker-supplied keypath metadata, not from any proof that we can/will sign the input. A presigned foreign input with a forged (zero-xfp) keypath sails through and the device still exports a signed POR committing to the foreign prevouts via SIGHASH_ALL.

Also worth noting: the regression test only models the weakest variant (a foreign-looking input made by stripping keypaths from one of the wallet's own UTXOs). There is no test with a genuinely foreign UTXO (different seed), nor for the presigned-foreign-input bypass described inline.

Recommend strengthening the condition to `not inp.required_key` and adding a test for the presigned-foreign-input variant before merge.

shared/psbt.py
scgbckbone: Addressed in 8dba7380. The gate at psbt.py:1888 now checks `not inp.required_key` instead of `num_our_keys`, so keypath metadata is no longer trusted: forged zero-xfp keypaths never survive to signing, and presigned inputs (fake keypath + partial sig) resolve `required_key = None` and are hard-rejected. Regression test `test_bip322_por_presigned_foreign_input` added (foreign UTXO from a different seed, zero-xfp keypath, garbage partial sig) — verified red without the fix and green with it; full `test_bip322.py` passes on both PSBT v0 (101) and v2 (98).

PR #774, opened 2026-09-01
bugfix: restrict HSM BIP322 WIF signing

scgbckbone:
## Summary

- Require msg_paths = ["any"] for BIP-322 signing with WIF Store keys.
- Preserve derivation-path policy matching for seed-backed keys.
- Add a regression test for forged BIP32 metadata on a WIF-backed P2WPKH input.

This is the master-specific backport, adapted to the older PSBT representation. Taproot is intentionally not included because master does not support Taproot spending.

## Testing

- HSM BIP-322 message-policy tests, PSBTv0: 5 passed
- HSM BIP-322 message-policy tests, PSBTv2: 5 passed

The regression test was also run without the fix and failed because the restricted m/0 policy incorrectly authorized signing, confirming it covers the bypass.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -16,7 +16,8 @@ This lists the new changes that have not yet been published in a normal release.
 
 ## 5.6.x - 2026-0x-xx
 
-- tbd
+- Bugfix: Require unrestricted HSM message-signing policy when signing BIP-322
+  messages with WIF Store keys.
 
 
 # Q Specific Changes
--- a/shared/hsm.py
+++ b/shared/hsm.py
@@ -897,6 +897,12 @@ class HSMPolicy:
                         if not inp.required_key:
                             continue
 
+                        required_keys = inp.required_key if inp.is_multisig else [inp.required_key]
+                        if any(pk in psbt.wif_store for pk in required_keys):
+                            if 'any' not in self.msg_paths:
+                                raise ValueError("WIF Store message signing requires any path")
+                            continue
+
                         if inp.is_multisig:
                             paths = [
                                 keypath_to_str(inp.subpaths[pk])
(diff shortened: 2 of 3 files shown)


scgbckbone (commented):
Deep review of the fix and its surroundings (`HSMPolicy.approve_transaction`, `psbt.py` WIF-store signing path, `parse_subpaths`, `approve_msg_sign`, CCC, and the WIF menu flows).

The bug is real pre-fix: `inp.subpaths` is attacker-controlled PSBT metadata, and for WIF Store keys the claimed derivation path is meaningless — the device signs with the raw imported key regardless of path (psbt.py:2227-2228). A forged BIP32 derivation entry (real master XFP + `m/0`) on a WIF-backed input let a restricted `msg_paths` policy authorize BIP322 signing with any key in the WIF Store. Seed-backed keys were never affected because they pass `check_pubkey_at_path` at signing time.

- Every POR input whose `required_key` (single pubkey or multisig set) intersects `psbt.wif_store` now requires the `any` policy — the only honest policy for keys with no derivation path. Multisig sets are fully scanned via `any(...)`; mixed seed+WIF multisig is caught.
- Placement before the `subpaths` lookup also fixes the latent KeyError refusal for WIF inputs with no subpath metadata, enabling the intended `msg_paths=["any"]` flow.
- No remaining variant of the bug class on master: `approve_msg_sign` is seed-subpath-only, the WIF "Sign MSG" menu is physical-UX only, CCC has no BIP322 path, and exact 33-byte pubkey matching has no compressed/uncompressed confusion (the store only accepts compressed keys). Taproot is N/A on master.
- No false positives: seed-only BIP322 flows and non-POR HSM transactions take unchanged branches; `msg_paths=["any"]` WIF users keep working.
- The regression test builds the exact attack shape (forged `m/0` metadata with the real simulator master fingerprint on a WIF-backed P2WPKH input), provably fails without the fix, and its second leg verifies the allowed `any` flow still signs.

Nit (non-blocking): the `settings_remove("wifs")` cleanup at the end of the test won't run if an assertion fails mid-test; a `try/finally` would be tidier.

OK by KIMI 3

PR #775, opened 2026-09-01
bugfix: bind firmware upgrade approval to staged image

scgbckbone:
`FirmwareUpgradeRequest` renders consent from a header snapshot, but the final flash operation consumes a live PSRAM window. A second upload can overwrite that window while approval is pending.

**Fix:**

- Add a dedicated `upgrade_write_count`, separate from transaction signing.
- Capture it synchronously when `FirmwareUpgradeRequest` is created, alongside the approved header snapshot.
- Assert it is unchanged immediately before `bootrom_takeover()` and `pa.firmware_upgrade()`.
- Increment it for normal PSRAM writes and for the native VirtDisk copy that bypasses `PSRAMWrapper.write_at()`.

There is no await between the final counter check and the flash call, so another cooperative task cannot modify the staged image after verification. This avoids hashing the complete firmware image twice.

The regression test uploads valid firmware A with its trailer, overwrites PSRAM with different same-length valid firmware B through the raw upload path, confirms the displayed version is still A, approves, and asserts that `pa.firmware_upgrade()` was never reached.

Tests:

- `testing/test_upgrades.py`: 5 passed on Mk5 simulator
- `test_upgrade_staged_image_tamper`: passed on Q1 simulator
- `test_psram_write_counter_covers_txn_output_region`: passed

### DIFF
--- a/shared/psram.py
+++ b/shared/psram.py
@@ -15,6 +15,10 @@ class PSRAMWrapper:
     # mutation of the transaction being signed
     txn_write_count = 0
 
+    # kept separate from transaction signing: firmware approval captures this
+    # value and aborts if its staged image is overwritten
+    upgrade_write_count = 0
+
     def __init__(self):
         self._wr = uctypes.bytearray_at(self.base, self.length)
 
@@ -31,6 +35,7 @@ class PSRAMWrapper:
         if offset < 2 * version.MAX_TXN_LEN:
             # covers both TXN staging regions (input and output)
             self.txn_write_count += 1
+            self.upgrade_write_count += 1
 
         return memoryview(self._wr)[offset:offset+ln]
 
--- a/shared/psram.py
+++ b/shared/psram.py
@@ -2,7 +2,7 @@
 #
 # psram.py -- access PSRAM chip on Mk4
 #
-import version, uctypes
+import uctypes
 
 # already started and memory mapped by bootrom.
 
@@ -10,16 +10,8 @@ class PSRAMWrapper:
     base = 0x9000_0000     # OCTOSPI1
     length = 0x40_0000     # 4 meg (lower half)
 
-    # bumped on every write into the TXN staging regions (input below
-    # MAX_TXN_LEN, output at/above it); used to detect post-review
-    # mutation of the transaction being signed
-    txn_write_count = 0
-
-    # kept separate from transaction signing: firmware approval captures this
-    # value and aborts if its staged image is overwritten
-    upgrade_write_count = 0
-
     def __init__(self):
+        self.psram_write_count = 0
         self._wr = uctypes.bytearray_at(self.base, self.length)
 
     def read_at(self, offset, ln):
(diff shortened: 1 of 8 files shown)


scgbckbone (commented):
Deep review of the fix and its surroundings (`FirmwareUpgradeRequest`, all PSRAM writers, `psramdisk.c`, the USB/vdisk/MicroSD upgrade entry flows, and the simulator model).

The vulnerability is real: the consent screen renders from a header snapshot while `pa.firmware_upgrade()` flashes live PSRAM, and uploads are explicitly allow-listed while an approval is on screen, so a second raw upload could swap in a different (validly signed — e.g. older, vulnerable) image while the user reads the prompt.

- I enumerated every writer to the bottom half of PSRAM (the staging region): all Python writes funnel through `PSRAMWrapper.write_at()`/`write()`, which now bumps the counter, and the one native bypass (`VBLKDEV.copy_file`, used only by `VirtDisk.import_file`) is covered by the new bump in vdisk.py. Host SCSI writes are confined to the top half of PSRAM in C and physically cannot touch the staging area.
- Capture ordering is correct in all three entry flows (USB: `authorize_upgrade` runs synchronously after the last write inside `handle_upload`; vdisk: the bump precedes `psram_upgrade`; MicroSD: all writes complete before the request is constructed).
- No TOCTOU window: no `await` between the assert and the flash call, asyncio is cooperative, and C callbacks are non-preemptive. The counter is monotonic and cannot be reset or wrapped by an attacker.
- Asserts are live in production (frozen modules are compiled without `-O`, and the codebase already relies on `assert` for equivalent security checks).
- False positives are effectively impossible: no legitimate flow writes PSRAM while the consent screen is displayed (Q1 LCD is SPI with on-stack buffers; `stok` polling, downloads, and `sha256` are read-only).
- The regression test reproduces the exact attack (upload A with trailer, overwrite with raw B, approve the stale prompt) and provably fails without the fix.

Nits (non-blocking): (1) `upgrade_write_count` is bumped inside the TXN-staging-region condition in psram.py — correct today only because `2 * MAX_TXN_LEN` equals the full PSRAM length on both boards, so a comment noting the coupling (or hoisting the bump) would future-proof it; (2) when the tamper check fires the prompt silently disappears — an on-device "upload changed, upgrade aborted" story would make the security event visible to the local user; (3) no positive-path test (approve an untampered image) exists, so a capture-ordering regression that bricks all upgrades would go undetected, and the vdisk.py bump is unreachable in the simulator because `SimulatedVirtDisk.import_file` overrides the method.

OK by KIMI 3

shared/psram.py
```diff
@@ -15,6 +15,10 @@ class PSRAMWrapper:
     # mutation of the transaction being signed
     txn_write_count = 0
 
+    # kept separate from transaction signing: firmware approval captures this
+    # value and aborts if its staged image is overwritten
+    upgrade_write_count = 0
```
doc-hex: need a `self.` here?

doc-hex:
do we need both write counters? couldn't they be a "psram_write_count" instead?

PR #776, opened 2026-09-01
bugfix: vdisk import bumps txn write count

scgbckbone:
The C VirtDisk importer (`ckcc.PSRAM.copy_file`, used only by `VirtDisk.import_file`) memcpy's into the PSRAM TXN staging regions without going through `PSRAMWrapper.write_at()`, so `txn_write_count` is not bumped. The pre-signing re-hash is gated on that counter, allowing a Virtual Disk firmware import to replace a reviewed PSBT while the guard takes its fast path.

**Fix:** bump the counter and revoke any PSRAM lease (`ALLOWED_DOWNLOAD`) in `VirtDisk.import_file` before calling `copy_file`. Bumping first also invalidates approval if the native copy fails after a partial write.

**Regression:** make the simulator model `copy_file` as the same direct lower-PSRAM write used on hardware, then drive the real auto-VirtDisk path: stage a PSBT, show its approval, drop a `.dfu` file, approve, and verify signing aborts with `Transaction modified`.

Precondition for exploit remains **Settings > Hardware On/Off > Virtual Disk = Enable & Auto** (`vidsk == 2`), not the shipped default.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -6,6 +6,8 @@ This lists the new changes that have not yet been published in a normal release.
 
 - Bugfix: Reject foreign inputs from BIP-322 Proof of Reserves, including inputs
   disguised with forged key-path metadata or partial signatures.
+- Bugfix: Detect and abort transaction signing if a Virtual Disk firmware import
+  overwrites the reviewed PSBT. Thanks to Huzaifa Jawaid.
 - Bugfix: Restore the ability to view the device-generated seed before adding user
   entropy, which was available in the previous dice-roll workflow but was inadvertently
   removed in 5.6.1/1.5.1Q. The new **View TRNG Words** menu item displays the full
--- a/shared/vdisk.py
+++ b/shared/vdisk.py
@@ -103,6 +103,12 @@ class VirtDisk:
         # copy file into another area of PSRAM where rest of system can use it
         assert sz <= MAX_UPLOAD_LEN       # too big
 
+        # C importer memcpy's into the TXN staging regions without going
+        # through the PSRAM wrapper — record it, and revoke any lease of
+        # those staged bytes, so post-review tampering can't hide
+        glob.ALLOWED_DOWNLOAD = None
+        glob.PSRAM.txn_write_count += 1
+
         # I could not resist doing this in C... since we already have the
         # data in memory, why mess around with file concepts?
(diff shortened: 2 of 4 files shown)


scgbckbone (commented):
Deep review of the fix and its surroundings (`VirtDisk.import_file`, `psram_copy_file` in psramdisk.c, the `ApproveTransaction` re-check in auth.py, and every other writer into the TXN staging regions).

The bug is real: `psram_copy_file` memcpy's directly into `PSRAM_BOT_BASE` — the live TXN staging region — bypassing `PSRAMWrapper.write_at()`, so a `.dfu` dropped onto an auto-consume virtual disk could overwrite a reviewed PSBT while the `txn_write_count`-gated tamper guard took its fast path. I verified `copy_file` is genuinely the only such path: host SCSI writes are range-checked into the top half only, `psram_wipe_and_setup` touches only the top half, Q1's psramdisk.c is a symlink to MK4's, and all other staging paths (MicroSD upgrade, vdisk PSBT via `SFFile`, USB/NFC/BBQr/QR/teleport) go through the counting wrapper.

- Revoking `ALLOWED_DOWNLOAD` and bumping `txn_write_count` *before* the C copy is fail-closed: if `copy_file` raises after a partial write, approval is already invalidated. Revoking the lease is also correct because a >2MB import can clobber the output region that may be leased for `dwld`.
- The bump can never cause a false rejection of an unmodified transaction: the auth re-check re-hashes and adopts the new baseline when bytes are unchanged (auth.py:619-626).
- No legitimate flow breaks: normal vdisk firmware upgrades still work (the bump is harmless when no signing is active), and failing a signing whose staged bytes were actually overwritten is the intended behavior, including on the HSM path which shares the same re-check.
- The simulator change is essential, not cosmetic: the old sim `import_file` override used the *counting* `PSRAM.write_at`, so the simulator could never reproduce the hardware bug; the new `SimBlockDev.copy_file` models the native bypass via `PSRAM._wr`, letting the shared vdisk.py code under test run unmodified. The regression test fails at two distinct assertion points without the fix, and the `pytest.raises(CCProtoError, match='Transaction modified')` assertion additionally proves fail-closed behavior.

Nit (non-blocking): the new test does `settings_set('vidsk', 2)` but omits the `sim_exec` guard instantiating `VirtDisk()` that both sibling tests use — fine under the official runner (which launches with `--set vidsk=1`), but a standalone run against a simulator started without that flag fails misleadingly.

OK by KIMI 3

PR #777, opened 2026-09-01
bugfix: reject incomplete wrapped segwit inputs

scgbckbone:
## Summary

- reject P2SH-P2WSH inputs that omit their redeem script
- keep malformed inputs fail-closed; no redeem script is derived or synthesized
- cover the reported high-fee malformed PSBT with a focused regression test

## Testing

- missing-redeem regression (PSBTv0 and PSBTv2): 2 passed
- sparse and wrapped WIF Store paths (PSBTv0 and PSBTv2): 8 passed
- adjacent witness-UTXO policy tests: 3 passed
- ordinary 1-of-1 multisig signing: 2 passed
- `git diff --check`

### DIFF
--- a/shared/psbt.py
+++ b/shared/psbt.py
@@ -836,8 +836,12 @@ class psbtInputProxy(psbtProxy):
 
                 redeem_script = self.get(ks)
 
-            if self.addr_fmt == AF_P2SH and self.witness_script and not self.redeem_script:
-                raise FatalPSBTIssue('Missing redeem script for input #%d' % my_idx)
+            if self.addr_fmt == AF_P2SH and self.witness_script:
+                expect = b'\x00\x20' + ngu.hash.sha256s(self.get(self.witness_script))
+                if not self.redeem_script or self.get(self.redeem_script) != expect:
+                    # redeem script is the only proof the UTXO is really segwit;
+                    # without it the input amount (and fee) cannot be verified
+                    raise FatalPSBTIssue('Missing/bad redeem script for input #%d' % my_idx)
 
             self.scriptSig = redeem_script
 
--- a/shared/psbt.py
+++ b/shared/psbt.py
@@ -836,6 +836,9 @@ class psbtInputProxy(psbtProxy):
 
                 redeem_script = self.get(ks)
 
+            if self.addr_fmt == AF_P2SH and self.witness_script and not self.redeem_script:
+                raise FatalPSBTIssue('Missing redeem script for input #%d' % my_idx)
+
             self.scriptSig = redeem_script
 
             if not addr_is_segwit and len(redeem_script) == 22 and \
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -8,6 +8,8 @@ This lists the new changes that have not yet been published in a normal release.
   disguised with forged key-path metadata or partial signatures.
 - Bugfix: Detect and abort transaction signing if a Virtual Disk firmware import
   overwrites the reviewed PSBT. Thanks to Huzaifa Jawaid.
+- Bugfix: Reject malformed PSBTs containing P2SH-P2WSH inputs without a redeem
+  script, preventing transactions with an unknown fee from proceeding to approval.
 - Bugfix: Restore the ability to view the device-generated seed before adding user
   entropy, which was available in the previous dice-roll workflow but was inadvertently
   removed in 5.6.1/1.5.1Q. The new **View TRNG Words** menu item displays the full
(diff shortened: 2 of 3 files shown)


shared/psbt.py
```diff
@@ -836,6 +836,9 @@ def determine_my_signing_key(self, my_idx, utxo, my_xfp, psbt, cosign_xfp=None):
 
                 redeem_script = self.get(ks)
 
+            if self.addr_fmt == AF_P2SH and self.witness_script and not self.redeem_script:
```
scgbckbone: **Major: incomplete fix — a *present but incorrect* redeem script bypasses this check and keeps the fee unverifiable.**

Craft a 2-of-3 P2SH-P2WSH input with a real `witness_utxo` (real spk, real amount), the real `witness_script`, and a ≥22-byte junk `redeem_script`:

1. `validate()` (psbt.py:665-666) only asserts the field's length, and this new check passes because the field is truthy.
2. `ks = self.witness_script or self.redeem_script` (psbt.py:833) prefers the witness script, so all downstream input validation — `validate_script()` and `target_spk == utxo.scriptPubKey` (psbt.py:933-936) — runs against the *witness* script and passes for the real UTXO. The PSBT-level `redeem_script` is never validated against the witness script on the input side.
3. But `witness_utxo_is_provably_segwit()` (psbt.py:761-773) uses the junk `redeem_script` (its `hash160(redeem_script) == addr_or_pubkey` test fails), so the input lands in `unverified_witness_utxo` (psbt.py:1881) → `total_value_in = None` (psbt.py:1933) → `calculate_fee()` returns None → the fee-limit check is skipped entirely (psbt.py:1679), leaving only the "Unable to calculate fee" warning (psbt.py:1938) — exactly the pre-fix state this PR exists to eliminate.
4. The input is still marked `AF_P2WSH_P2SH`/segwit and signed with a valid BIP143 digest committing the real amount. The emitted partial signatures are valid for the real on-chain UTXO; the attacker reassembles a valid tx with the correct scriptSig, so the broken scriptSig the device would emit itself is no defense. `history.verify_amount` silently records first-seen outpoints and is no backstop either.

Suggested fix — when `witness_script` is present, require the redeem script to be present *and* correct:

```python
if self.addr_fmt == AF_P2SH and self.witness_script:
    if not self.redeem_script or self.get(self.redeem_script) != \
            b'\x00\x20' + ngu.hash.sha256s(self.get(self.witness_script)):
        raise FatalPSBTIssue('Missing/bad redeem script for input #%d' % my_idx)
```

More robustly: make the `unverified_witness_utxo` condition fatal for any input we actually sign (`required_key` set) instead of a warning. A companion test with `redeem_script = b'\x00\x20' + b'\x11'*32` would pin this down — it currently reaches the approval UX instead of being rejected.

scgbckbone (commented):
Deep review of the fix and its surroundings (`determine_my_signing_key`, `witness_utxo_is_provably_segwit`, `consider_inputs`, `calculate_fee`, finalize path).

What checks out: the bug is real and correctly targeted — pre-fix, a P2SH-P2WSH input with `witness_utxo` + `witness_script` but no `redeem_script` was treated as segwit for signing while `witness_utxo_is_provably_segwit()` returned False, making the fee uncomputable and degrading the fee-limit check to a warning before approval. The new raise fires inside `consider_inputs`, which is on every signing path (SD/NFC/QR/USB, HSM, CCC, teleport, BIP322 PoR). No false positives: BIP174 mandates the redeem script for wrapped-segwit inputs so compliant producers are unaffected, foreign/presigned inputs return early before the check, native P2WSH skips it, legacy P2SH is untouched, and singlesig P2SH-P2WPKH was already covered. The regression test is well-constructed — pre-fix the flow proceeds to the approval story and the test goes red; the asserted message matches verbatim.

However, the fix is incomplete: the identical "unknown fee" hole remains open when the redeem script is *present but bogus* — see the inline comment for the full trace. The exploit is a trivial adaptation of the one being fixed (swap "omit redeem script" for "supply garbage redeem script"), with the same impact. The test also doesn't pin down this remaining variant.

Recommend extending the check to validate the redeem script's contents (or making unverifiable own-input witness UTXOs fatal) before release, or at minimum tracking a follow-up.

shared/psbt.py
scgbckbone: Addressed in 26d8aab9. The check now validates the redeem script contents: for P2SH inputs carrying a witness script it must equal `OP_0 <sha256(witness_script)>` (same idiom as the change-output side at psbt.py:544), so a present-but-wrong redeem script is hard-rejected like the missing case — the signing-side segwit decision and the fee-side provably-segwit decision can no longer disagree. Regression test `test_junk_p2sh_p2wsh_redeem_script_rejected` added: verified to reach the `OK TO SEND?` approval UX with only the unverified-witness-UTXO warning without the fix, and to raise with it. Over-rejection sweep green: 1-of-1 multisig, WIF-store multisig/wrapped signing (all addr formats), sh-vs-wrapped PSBTs, and a full 2-of-3 P2SH-P2WSH `test_ms_sign_simple` all still pass.

PR #778, opened 2026-09-01
bugfix: harden temporary wallet write guards

scgbckbone:
## Summary

- require that no master or temporary secret is active before permanent XPRV import
- apply the same precondition to permanent TAPSIGNER backup import
- use the active-secret check for developer and CKCC/USB backup restores
- use the active-secret check when receiving Key Teleport secret material
- use the active-secret check for Seed XOR completion and its temporary-seed warning
- reject low-level master seed writes and full master restores while a temporary wallet is active
- document why checking only whether the secure-element slot is blank is insufficient
- add one changelog entry describing the affected empty-master plus temporary-seed state
- add a seedless regression test covering both developer and CKCC backup restore prompts with a temporary seed active

No Main PIN changes are included as those are opened here https://github.com/Coldcard/firmware/pull/767

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -11,6 +11,7 @@ This lists the new changes that have not yet been published in a normal release.
   verification of dice-roll or coin-flip mixing.
 - Bugfix: Simulator crashed on Bless Firmware, due to a desynced LED pipe. Thanks to
   [@hitechhayekian](https://github.com/hitechhayekian).
+- Bugfix: With an empty master wallet and an active temporary seed, keep imports and backup restores temporary instead of treating them as master-seed changes.
 
 # Mk Specific Changes
 
--- a/shared/actions.py
+++ b/shared/actions.py
@@ -1478,7 +1478,9 @@ async def restore_backup_dev(*a):
     if fn:
         words = False if fn[-3:] == ".7z" else None
         import backups
-        await backups.restore_complete(fn, not pa.is_secret_blank(), words)
+        # A temporary wallet makes master writes no-op, despite the restore reporting success.
+        # Route the backup to another temporary wallet whenever any secret is active.
+        await backups.restore_complete(fn, pa.has_secrets(), words)
 
 async def bkpw_override(*A):
     # allows user to:
--- a/shared/actions.py
+++ b/shared/actions.py
@@ -1409,7 +1409,9 @@ async def import_xprv(_1, _2, item):
 
     ephemeral = item.arg
     if not ephemeral:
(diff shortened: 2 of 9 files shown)


scgbckbone (commented):
Deep review of the fix: I traced every changed call site end-to-end and enumerated all writers of the master secret.

The bug is real: `pa.change()` silently no-ops when a temporary seed is active (pincodes.py:355-356), and on a seedless device (SE secret blank) with an active temporary seed, several flows decided "master vs. temporary" using `pa.is_secret_blank()`/`has_se_secrets()`, which ignore `tmp_value`. Those flows would attempt a master-seed write that silently did nothing while the UI reported success — after reboot the wallet the user believed they imported is gone. I confirmed pre-fix reachability of all four buggy decision points (USB `rest` restore, dev-menu restore, Key Teleport receive `s`/`x`/`v`, Seed XOR restore).

- I enumerated every master-secret writer: `pa.change(new_secret=...)` is called only from `restore_from_dict_ll` (now guarded by the new assert), `set_seed_value` (now guarded), and `convert_ephemeral_to_master` (intentionally `tmp_lockdown=True` — the legitimate commit-tmp-to-SE path, unaffected). All routing decisions feeding those writers now use `pa.has_secrets()`, which correctly accounts for `tmp_value`.
- The two new `assert not pa.tmp_value` backstops are unreachable via legitimate flows: `set_seed_value`'s callers are all either guarded or only reachable from `EmptyWallet` (shown only when `not pa.has_secrets()`), and the menu stack is rebuilt after tmp activation, so there is no stale-menu bypass. I also checked QR-scan import, NFC import, BIP-85, seed vault, HSM, and trick-pin "Look Blank" mode (which zeroes the SE secret flag without setting `tmp_value`, so it neither triggers the asserts nor the bug) — none bypass the guards.
- No false positives: behavior is unchanged in all three non-buggy states (master present; blank SE without tmp; force-tmp), and existing direct callers of `restore_from_dict_ll` (test_se2.py, devtest) run without `tmp_value` set, so the new asserts don't break them.
- The new simulator test genuinely exercises both fixed routing paths on a seedless simulator and asserts "temporary" wording at both prompts — it would fail on unfixed code at both assertions.

Nits (non-blocking): (1) `from flow import has_secrets` inside the `dtype == 'b'` branch in teleport.py is now redundant with the function-level import; (2) `pa.has_secrets()` returns a truthy `bytes` object rather than a strict `bool` when SE is blank + tmp active — only truthiness is consumed downstream, and this matches existing codebase idiom, but `bool(...)` would be cleaner.

OK by KIMI 3

PR #788, opened 2026-09-02
Harden and expand SegWit outpoint value caching

scgbckbone:


shared/history.py
```diff
@@ -116,28 +103,49 @@ def verify_amount(cls, prevout, amount, in_idx):
...
+
+        for in_idx, txin in psbt.input_iter():
+            inp = psbt.inputs[in_idx]
+            if inp.is_segwit and inp.added_sigs and inp.required_key and inp.has_utxo():
+                if not cls.verify_amount(txin.prevout, inp.amount, in_idx, cache):
+                    # add() serializes prevout immediately, which is important because
+                    # PSBTv0 input_iter() reuses and mutates its CTxIn object
+                    cls.add(cache, txin.prevout, inp.amount)
+                    changed = True
+
```
doc-hex, line 130: "change" outputs of the TXN could be captured here as well.

doc-hex: but that interacts with add_segwit_utxos_finalize

scgbckbone: you need txid for it so only if we move after txn is finalized - now checking if we can/should unify ...

scgbckbone: I would keep them separate because they become trustworthy at different
  moments:

  - Input amounts should be committed immediately after Coldcard produces
    signatures. This must also work for partially signed multisig PSBTs that
    Coldcard cannot finalize.

  - Change outputs should be committed only after finalization, when the final
    txid is known.

doc-hex: that makes sense, good.

PR #795, opened 2026-09-03
EDGE: Use one PSRAM write counter

scgbckbone:
`FirmwareUpgradeRequest` renders consent from a header snapshot, but the final flash operation consumes a live PSRAM window. A second upload can overwrite that window while approval is pending.

**Fix:**

- Add a dedicated `upgrade_write_count`, separate from transaction signing.
- Capture it synchronously when `FirmwareUpgradeRequest` is created, alongside the approved header snapshot.
- Assert it is unchanged immediately before `bootrom_takeover()` and `pa.firmware_upgrade()`.
- Increment it for normal PSRAM writes and for the native VirtDisk copy that bypasses `PSRAMWrapper.write_at()`.

There is no await between the final counter check and the flash call, so another cooperative task cannot modify the staged image after verification. This avoids hashing the complete firmware image twice.

The regression test uploads valid firmware A with its trailer, overwrites PSRAM with different same-length valid firmware B through the raw upload path, confirms the displayed version is still A, approves, and asserts that `pa.firmware_upgrade()` was never reached.

Tests:

- `testing/test_upgrades.py`: 5 passed on Mk5 simulator
- `test_upgrade_staged_image_tamper`: passed on Q1 simulator
- `test_psram_write_counter_covers_txn_output_region`: passed

### DIFF
--- a/shared/auth.py
+++ b/shared/auth.py
@@ -400,7 +400,7 @@ class ApproveTransaction(UserAuthorizedAction):
         # - they are re-read from live PSRAM during display, signing & finalization,
         #   and a USB host could rewrite them while we wait for approval
         from glob import PSRAM
-        self.parsed_write_count = PSRAM.txn_write_count
+        self.parsed_write_count = PSRAM.psram_write_count
         self.parsed_sha = psram_sha256(self.offset, self.psbt_len)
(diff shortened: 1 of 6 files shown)


scgbckbone (commented):
OK by KIMI 3

PR #796, opened 2026-09-03
EDGE: Validate wrapped-segwit redeem script contents

scgbckbone:
## Summary

- reject P2SH-P2WSH inputs that omit their redeem script
- keep malformed inputs fail-closed; no redeem script is derived or synthesized
- cover the reported high-fee malformed PSBT with a focused regression test

## Testing

- missing-redeem regression (PSBTv0 and PSBTv2): 2 passed
- sparse and wrapped WIF Store paths (PSBTv0 and PSBTv2): 8 passed
- adjacent witness-UTXO policy tests: 3 passed
- ordinary 1-of-1 multisig signing: 2 passed
- `git diff --check`


scgbckbone (commented):
OK by KIMI 3

PR #798, opened 2026-09-03
EDGE: usb: ncry v3

scgbckbone:
## Summary

Implements opt-in USB ncry v3 authenticated encryption. V3 derives independent AES-CTR and HMAC-SHA256 keys for each direction and authenticates every frame over its direction, sequence number, length, and ciphertext. This rejects tampering, replay, reordering, and cross-direction reflection; authentication or framing failures terminate the session.

The default remains ncry v1, with v1/v2 behavior covered by regression tests. V3 protects the encrypted channel but does not authenticate the COLDCARD endpoint, so clients requiring endpoint authentication must use `check_mitm()` with a previously trusted xpub.

Includes matching ckcc-protocol source version `1.6.0` support, fixed test vectors, and protocol documentation. The submodule points at merged client commit `c39db297963bf2ccc592d1ee287e7ae2c6b613c8`, including the follow-up transport and response-limit fixes.

## Release target

- ncry v3 was not included in the published Mk4/Mk5 5.6.1 or Q 1.5.1Q firmware
- firmware support remains unreleased and is planned for the next standard Mk4/Mk5 and Q releases; exact version numbers are intentionally left open in `Next-ChangeLog.md`

## Validation

- merged current firmware master through the final Mk4/Mk5 release commit `e33ef9228b023a1df43eb4b41bff768234b08e65`
- `git diff origin/master --check`
- ckcc-protocol ncry v3 unit tests: **12 passed, 4 hardware tests deselected**
- changed Python files compile under CPython 3.11
- the existing PR review record includes **12/12** firmware simulator tests; the master merge did not overlap the ncry implementation or its simulator tests

### DIFF
--- a/shared/usb.py
+++ b/shared/usb.py
@@ -5,11 +5,14 @@
 import ckcc, pyb, callgate, sys, ux, ngu, stash, aes256ctr, ujson
 from uasyncio import sleep_ms, core
 from uhashlib import sha256
-from public_constants import MAX_MSG_LEN, MAX_BLK_LEN
-from public_constants import STXN_FLAGS_MASK
+from public_constants import (
+    MAX_BLK_LEN, MAX_MSG_LEN, STXN_FLAGS_MASK,
+    USB_V3_C2D, USB_V3_D2C, USB_V3_KDF_LABEL,
+    USB_V3_MAX_WIRE_MSG_LEN, USB_V3_TAG_LEN,
+)
 from ustruct import pack, unpack_from
 from ckcc import watchpoint, is_simulator
-from utils import problem_file_line, call_later_ms
+from utils import problem_file_line, call_later_ms, consteq, hkdf_expand
 from version import supports_hsm, is_devmode, MAX_TXN_LEN, MAX_UPLOAD_LEN
 from exceptions import FramingError, CCBusyError, HSMDenied, HSMCMDDisabled, SpendPolicyViolation
 from pincodes import pa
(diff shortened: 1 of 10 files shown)


shared/usb.py
```diff
@@ -280,9 +323,24 @@ async def usb_hid_recv(self):
                 reason = exc.args[0]
                 # print("Framing: %s" % reason)
                 self.reset_upload()
-                await self.framing_error(reason)
+                if self.ncry_ver == 0x3:
+                    try:
+                        await self.framing_error(reason)
```
scgbckbone, line 328: In a v3 session this fram reply can go out unencrypted: send_response encrypts only when encrypted_req was set, but FramingError was raised earlier (badsz/xlong/must encrypt/no key/auth). Encrypt fram once v3 keys exist (session-level bound) instead of per-request encrypted_req, so the reason string does not leak.

scgbckbone (commented):
Low finding (would be REQUEST_CHANGES, but review is on own PR — GitHub blocks that): one plaintext fram frame can be emitted in an otherwise always-encrypted v3 session. A malformed first v3 packet raises FramingError before encrypted_req is set, so send_response goes out unencrypted. Client latches 'Unencrypted response in ncry v3' and kills the session; only the reason string leaks. Suggest encrypting fram once v3 keys exist (session-level bound), not per-request flag. NOTE: sim tests not runnable locally (missing SDL2 for simulator python), so this is static + prior full-file review.

PR #799, opened 2026-09-03
EDGE: Reject non-ASCII BIP-39 passphrases at entry points

scgbckbone:
Require BIP-39 passphrases to use printable ASCII only (`0x20–0x7e`). Reject Unicode, tabs, and newlines across USB, saved-passphrase, QR, and Notes flows.


scgbckbone (commented):
OK by KIMI 3

scgbckbone (commented):
OK by KIMI 3

PR #800, opened 2026-09-03
EDGE: Wipe on duress-secret view/activate in Delta Mode

scgbckbone:
## Problem

In Delta Mode, the Trick PINs management menu let a coercer (logged in with the delta PIN) reach live duress secrets without any wipe:

- **`duress_details`** (`shared/trick_pins.py:888-911`): "Press (6) to view associated secrets" reads the duress secret from the SE2 slot and renders the **full duress seed words / XPRV** on screen.
- **`activate_wallet`** (`shared/trick_pins.py:798-832`): loads the duress secret as the active temporary seed via `set_ephemeral_seed`.

Every other secret-revealing flow calls `wipe_if_deltamode()` (pwsave, teleport, Seed Vault, XOR restore, BIP-85, backups); these two paths were missed. The **master seed itself is safe** (duress wallets are hardened BIP-32 children), but the duress wallet's funds are exposed and the user's deniability collapses (the menu also enumerates trick-PIN policies).

## Fix

Keep menu browsing working — Delta Mode should look like normal operation, so we do **not** wipe on opening the menu — but call `wipe_if_deltamode()` at the two secret-revealing actions:

- before the (6) viewer reads the slot (`duress_details`)
- before `activate_wallet` applies the secret

Minimal diff: 2 guard calls in `shared/trick_pins.py`.

## Test

- **New** `test_deltamode_duress_secret_view_wipes`: browsing the Trick PINs menu works in delta mode, but viewing the duress secret via (6) and activating the wallet each trigger a wipe (asserted via a `fast_wipe` spy, since a real wipe reboots the sim). **Verified to fail pre-patch** (no wipe) and pass with the fix.
- Broader `test_se2.py` duress/deltamode/trick subset: 38 passed; the one failure (`test_se2_trick_backups`) is **pre-existing** (confirmed failing on unpatched code — unrelated backup-length assertion).

## Changelog

Added a `Next-ChangeLog.md` entry.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -33,6 +33,9 @@ This lists the new changes that have not yet been published in a normal release.
 - Bugfix: Reject non-ASCII BIP-39 passphrases (USB, saved-passphrase recall, and
   note/password lanes) instead of silently deriving a wallet incompatible with
   BIP-39-normalizing software.
+- Bugfix: In Delta Mode, wipe the seed if anyone tries to view or activate a duress
+  wallet's secret from the Trick PINs menu, instead of revealing it. Browsing the menu
+  itself still works, so Delta Mode continues to look like normal operation.
 
 # Mk Specific Changes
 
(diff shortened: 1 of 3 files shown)


scgbckbone (commented):
OK by KIMI 3

PR #801, opened 2026-09-03
EDGE: Fix SSSP block height reset DoS

scgbckbone:
An approved SSSP PSBT can persist an arbitrarily large height-like nLockTime in `block_h`, causing later legitimate transactions to fail with rewound or velocity violations.

Add the block-height recovery action already available in CCC's Last Violation screen. The reset remains behind SSSP's existing non-hobbled configuration gate: the bypass PIN and, when enabled, the first/last seed-word challenge. This closes the recovery asymmetry between CCC and SSSP.

Tested on the Mk5 and Q1 simulators.


scgbckbone (commented):
OK by KIMI 3

PR #803, opened 2026-09-03
EDGE: bugfix: bound firmware to world checksum

scgbckbone:
Reject firmware past the world-checksum-covered flash boundary and align the signing limit.

Tested: Mk4 and Q1 verifier builds.


scgbckbone (commented):
OK by KIMI 3

PR #805, opened 2026-09-03
add proper error for firmware re-staging

scgbckbone:


shared/actions.py
```diff
@@ -2089,6 +2089,7 @@ async def main_pin_changer(*a):
     # - the bootloader maybe lying to us about main vs trick pin
     # - what may look like just policy here, is in fact enforced by the bootrom code
     #
+    assert not pa.tmp_value      # menu already hides this; tmp wallet has no main PIN
```
scgbckbone, line 2092: just defense in depth - this menu item is not shown to tge user if in tmp mode

testing/test_ccc.py
```diff
@@ -1010,9 +1010,6 @@ def test_maxed_out(settings_set, setup_ccc, enter_enabled_ccc, ccc_ms_setup, sim
     import_multisig(data=ms_conf)
     press_select()  # confirm multisig import
 
-    # get rid of last violation - as it is held as global
-    sim_exec('from ccc import CCCFeature; CCCFeature.last_fail_reason=""')
```
scgbckbone, line 1014: lfr is no longer class attribute - but persisted setting

testing/test_ccc.py
```diff
@@ -1135,8 +1132,6 @@ def test_load_and_sign_key_C(settings_set, setup_ccc, enter_enabled_ccc, ccc_ms_
...
```
scgbckbone, line 1139: lfr is no longer class attribute - but persisted setting

shared/auth.py
```diff
@@ -1609,7 +1609,9 @@ async def interact(self):
             ch = await ux_show_story(msg)
 
             if ch == 'y':
-                assert glob.PSRAM.psram_write_count == self.psram_write_count
+                if glob.PSRAM.psram_write_count != self.psram_write_count:
+                    # staged bytes changed since approval prompt was shown
+                    return await self.failure("Firmware modified")
```
scgbckbone, line 1614: assert was caught by general eexception handler and failure was silent - now changed to have error message

PR #807, opened 2026-09-03
psbt: unified opt-in signature hash

kwsantiago:
Bitcoin Knots `v29.4.1.knots20260508` defines one signature message for every script type, opted into per input by hash type bit `0x20`. It is specified in [doc/unified-sighash.md](https://github.com/bitcoinknots/bitcoin/blob/v29.4.1.knots20260508/doc/unified-sighash.md) at that tag.

That release is a hardfork of Bitcoin and also changes its proof of work. This is the signature message alone. The COLDCARD is a signer: it signs the hash type the PSBT declares and carries no activation height, so deciding when to opt in stays with the wallet.

## Adds

- `psbtObject.make_txn_unified_sighash()`, covering bare/P2SH and segwit v0.
- The six opted-in hash types in `ALL_SIGHASH_FLAGS`. `0x20` and `0xa0` name no output type and are refused.
- An approval note saying what the signature covers, and the hash type on the per-input screen.

A signature that opts in commits to every spent amount and scriptPubKey rather than only the input being signed, closing CVE-2020-14199 and CVE-2013-2292 for those inputs, and does not verify under the pre-fork rules, so the transaction cannot be replayed onto a chain without the rule.

ANYONECANPAY gets its own approval wording: those commit to that input's own spent output and to no other, so the note that fits the rest would be false for them.

Taproot is out of scope. This tree does not sign taproot, so the rule's script types 2 and 3 are unreachable and refused.

## Unchanged without the bit

The digest is selected by the declared hash type, and the extra work in `consider_inputs()` is gated on it, so a PSBT that never sets `0x20` takes master's path. 108 of 108 multisig tutorial cases pass across every legacy hash type, script type and M-of-N shape.

No existing test matrix grew: the opted-in types live in their own map rather than in `SIGHASH_MAP`.

## Verified

- 142 of the 166 cross-implementation vectors from Knots' [`src/test/data/unified_sighash.json`](https://github.com/bitcoinknots/bitcoin/blob/v29.4.1.knots20260508/src/test/data/unified_sighash.json), vendored byte-identical; the 24 taproot vectors are skipped and the count is printed.
- Accepted and mined by a Knots regtest node. The legacy digest wearing the opt-in byte is rejected, as it must be.
- CVE-2020-14199 reproduced: BIP-143 accepts a lied-about sibling amount, the opt-in rejects it.
- New tests skip when `bitcoind` has no such deployment, so a stock node runs the suite unchanged.

Depends on #806. The tests here import multisig wallets from the card, which needs that fixture fix, so its commit is included below until #806 lands; after that this reduces to the two commits that carry the feature.


nvk:
We only support bitcoin.

PR #808, opened 2026-09-03
Fix uncaught AttributeError on non-object JSON in parse_msg_sign_request

Amiga500:
parse_msg_sign_request() assumed any JSON-parseable input is a dict. Input of 123, "str", null, or [1,2] parses fine but .get() raises AttributeError, which escapes the except ValueError fallback and propagates past approve_msg_sign (catches only AssertionError, ValueError) to the fatal handler → device crash via NFC / QR / SD .json file.

Fix: raise ValueError for non-dict JSON so it falls through to the existing line-based parsing path.

data_dict = ujson.loads(data.strip())
if not isinstance(data_dict, dict):
    # valid JSON, but not an object (e.g. 123, "str", null, [1,2])
    raise ValueError("not a JSON object")
text = data_dict.get("msg", None)

Behavior for valid JSON objects, dict-without-msg, and plain-text inputs is unchanged.

### DIFF
--- a/shared/msgsign.py
+++ b/shared/msgsign.py
@@ -312,6 +312,9 @@ def parse_msg_sign_request(data):
 
     try:
         data_dict = ujson.loads(data.strip())
+        if not isinstance(data_dict, dict):
+            # valid JSON, but not an object (e.g. 123, "str", null, [1,2])
+            raise ValueError("not a JSON object")
         text = data_dict.get("msg", None)
         if text is None:
             raise AssertionError("MSG required")
--- a/shared/msgsign.py
+++ b/shared/msgsign.py
@@ -314,7 +314,7 @@ def parse_msg_sign_request(data):
         data_dict = ujson.loads(data.strip())
         if not isinstance(data_dict, dict):
             # valid JSON, but not an object (e.g. 123, "str", null, [1,2])
-            raise ValueError("not a JSON object")
+            raise ValueError
         text = data_dict.get("msg", None)
         if text is None:
             raise AssertionError("MSG required")
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -7,6 +7,9 @@ your addition and anything else already in this file.**
 
 # Shared Improvements - Both Mk and Q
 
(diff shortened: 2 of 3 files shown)


shared/msgsign.py
```diff
@@ -312,6 +312,9 @@ def parse_msg_sign_request(data):
 
     try:
         data_dict = ujson.loads(data.strip())
+        if not isinstance(data_dict, dict):
+            # valid JSON, but not an object (e.g. 123, "str", null, [1,2])
+            raise ValueError("not a JSON object")
```
doc-hex: better to just raise ValueError w/o the string as it saves flash space and will operate the same.

Amiga500: Done

scgbckbone (changes requested):
add this test to the end of `test_msg.py`

```python
@pytest.mark.parametrize('body,fail', [
    ('123', False),  # valid JSON, int
    ('"hello"', False),  # valid JSON, string
    ('null', False),  # valid JSON, null
    ('[1,2]', False),  # valid JSON, list
    # all above treated as just message
    ('{"nomsg": 1}', True),  # object but missing "msg" key
])
def test_sign_msg_json_not_object(body, fail, open_microsd, microsd_path, goto_home,
                                  pick_menu_item, cap_story):

    # valid JSON that is not an object (or lacks "msg") must produce a clean
    # failure story, never a crash (yikes)
    fname = 't-msgsign-bad.json'
    with open_microsd(fname, 'wt') as sd:
        sd.write(body)

    goto_home()
    pick_menu_item('Advanced/Tools')
    pick_menu_item('File Management')
    pick_menu_item('Sign Text File')
    time.sleep(.1)
    pick_menu_item(fname)
    time.sleep(.1)
    title, story = cap_story()
    if fail:
        assert not story.startswith('Ok to sign this?')
        assert 'MSG required' in story or 'must be ascii' in story or 'too short' in story
    else:
        assert story.startswith('Ok to sign this?')
```

Amiga500:
Done

scgbckbone (approved):
ACK https://github.com/Coldcard/firmware/pull/808/changes/fe526c2fab89df5c65dba1bfcc53a1c53d6c9a44

PR #809, opened 2026-09-03
Fix descriptor checksum bypass: reject descriptors with multiple '#' …

Amiga500:
checksum_check() used split("#") without maxsplit: two or more # → ValueError("too many values to unpack") → swallowed by the same handler as "no checksum present" → raw input returned unverified under csum_required=False.

Fix: split("#", 1) + reject any # remaining in the checksum portion.

desc, checksum = desc_w_checksum.split("#", 1)
...
if "#" in checksum:
    # never allow multiple '#' to bypass checksum verification
    raise ValueError("Too many '#' in descriptor")

Well-formed inputs (no checksum, or exactly one # with valid checksum) are unchanged; desc#checksum#junk now raises instead of bypassing.

### DIFF
--- a/releases/Next-ChangeLog.md
+++ b/releases/Next-ChangeLog.md
@@ -7,6 +7,9 @@ your addition and anything else already in this file.**
 
 # Shared Improvements - Both Mk and Q
 
+- Bugfix: Reject descriptors with multiple `#` characters, which silently bypassed
+  checksum verification on import. Thanks to [@Amiga500](https://github.com/Amiga500).
+
 - tbd
 
 
--- a/shared/descriptor.py
+++ b/shared/descriptor.py
@@ -142,11 +142,14 @@ class Descriptor:
     @staticmethod
     def checksum_check(desc_w_checksum , csum_required=False):
         try:
-            desc, checksum = desc_w_checksum.split("#")
+            desc, checksum = desc_w_checksum.split("#", 1)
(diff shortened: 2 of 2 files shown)


doc-hex:
please add a one-liner to Next-ChangeLog.md with a link to your github

Amiga500:
Done

scgbckbone (commented):
No checksum bypass existed before this change — multi-`#` inputs were already fail-closed. `desc_w_checksum.split("#")` with no maxsplit returns 3+ parts for `desc#sum#extra`, so the unpack into two variables raises `ValueError`, which the `except ValueError` catches and returns the *entire original string* (hashes included) as the descriptor with `checksum=None`. That string then fails downstream parsing (the `#`s are not valid in a key expression, so `parse_key_orig_info` / descriptor parsing rejects the import) — the checksum is never "skipped" with an attacker-controlled suffix.

So the premise in the changelog entry ("silently bypassed checksum verification") doesn't hold. This is a readability/error-message improvement at best.

Amiga500:
You're right — I verified the behavior: split("#") on input with 3+ parts raises ValueError, the whole original string is returned with checksum=None, and downstream parsing rejects it. The checksum was never actually bypassed, so the premise in the changelog entry doesn't hold. Closing this as not-a-bug. Thanks for the careful analysis!

PR #811, opened 2026-09-04
Docs: Fix ncry v3 protocol version references

scgbckbone:


docs/usb-ncry-v3.md
```diff
@@ -163,8 +163,9 @@ session.
 First supported in:
 
 - ckcc-protocol client: `1.6.0`
```
scgbckbone, line 165: we should release new ckcc (current 1.5.0)

doc-hex: done

PR #817, opened 2026-09-11
EDGE: Harden PSBTv2 parsing and preserve Miniscript context

scgbckbone:
## Summary

- reject key data on BIP370 singleton fields in global, input, and output maps
- require an explicit global version for PSBTv2 and reject non-exact global CompactSize counts
- preserve the Taproot context when deriving Miniscript nodes
- add focused parser and Miniscript regression coverage

### DIFF
--- a/releases/EdgeChangeLog.md
+++ b/releases/EdgeChangeLog.md
@@ -13,6 +13,7 @@ This lists the changes in the most recent EDGE firmware, for each hardware platf
 
 # Shared Improvements - Both Mk4 and Q
 
+- Enhancement: Support per-input required height and time locktimes in PSBTv2 transactions.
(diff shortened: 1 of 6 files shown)


releases/EdgeChangeLog.md
```diff
@@ -13,6 +13,7 @@ This lists the changes in the most recent EDGE firmware, for each hardware platf
 
 # Shared Improvements - Both Mk4 and Q
 
+- Enhancement: Support per-input required height and time locktimes in PSBTv2 transactions.
```
scgbckbone, line 16: this is from PR https://github.com/Coldcard/firmware/pull/816 where I forgot to add edge changelog entry

